PR #28, opened 2023-11-13
PO-46: Rename database to opal-fines-db

sabahirfan:
PO-46

### Change description ###
Rename database to opal-fines-db

```
[x] Yes
[] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -13,11 +13,11 @@ locals {
   db_name = "opal"
 }
 
-module "opal_db" {
+module "opal-fines-db" {
   providers = {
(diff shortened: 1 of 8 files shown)


infrastructure/main.tf
```diff
@@ -13,11 +13,11 @@ locals {
   db_name = "opal"
```
danlysiak, line 13: This needs to be updated too - this is what actually names the DB in higher environments

sabahirfan: done

RustyHMCTS (approved):
I'm not sure my approval means much, but APPROVED.

PR #36, opened 2023-11-15
Po 44 make get defendant account api query driven

DeclanClarkeCGI:
A change to the getDefendantAccount api to require query parameters rather than a JSON body to match front end requirements

[
](https://tools.hmcts.net/jira/browse/PO-44)

```
[ ] Yes
[ X] No
```


max-holland (approved):
Have pulled the branch and seems to be working as intended.

PR #142, opened 2024-01-19
Add some more tests for legacy data marshalling

RustyHMCTS:

```
[ ] Yes
[*] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/opal/dto/AccountSearchDto.java
+++ b/src/main/java/uk/gov/hmcts/opal/dto/AccountSearchDto.java
@@ -1,6 +1,7 @@
 package uk.gov.hmcts.opal.dto;
 
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import lombok.Builder;
 import lombok.Data;
 import java.util.Optional;
@@ -31,10 +32,10 @@ public class AccountSearchDto implements ToJsonString {
     /** Unsure. */
     private String tillNumber;
 
-    public Long getNumericCourt() {
+    @JsonIgnore
+    public Optional<Long> getNumericCourt() {
         return Optional.ofNullable(getCourt())
             .filter(s -> s.matches("[0-9]+"))
-            .map(Long::parseLong)
-            .orElse(null);
+            .map(Long::parseLong);
     }
 }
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/hmcts/opal/dto/AccountSearchDto.java
```diff
@@ -29,4 +30,11 @@ public class AccountSearchDto implements ToJsonString {
     private String majorCreditor;
     /** Unsure. */
     private String tillNumber;
+
+    public Long getNumericCourt() {
```
sabahirfan, line 34: How about  if we return `Optional<Long>` instead returning `null` t line 38 ?

RustyHMCTS: I did think about doing that, but in the places where this is being utilised, I actually want a null.

src/main/java/uk/gov/hmcts/opal/dto/legacy/DefendantAccountSearchCriteria.java
```diff
@@ -16,37 +17,50 @@
 @AllArgsConstructor
 public class DefendantAccountSearchCriteria implements ToJsonString {
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("account_number")
     private String accountNumber;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
```
sabahirfan, line 24: Is there an option to configure this on class level for all the fields?

RustyHMCTS: Yes - have done that in new commit

src/test/java/uk/gov/hmcts/opal/dto/legacy/DefendantAccountSearchCriteriaTest.java
```diff
@@ -0,0 +1,82 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static uk.gov.hmcts.opal.dto.legacy.DefendantAccountSearchCriteria.DefendantAccountSearchCriteriaBuilder;
+
+@Slf4j
+public class DefendantAccountSearchCriteriaTest {
+
+    @Test
+    public void testBuilder() throws InterruptedException {
```
sabahirfan, line 17: In my opinion we don't need to write unit tests for the generated code.

PR #188, opened 2024-02-08
Code Optimise and cleanup

sabahirfan:
### Change description ###
Code Optimise and cleanup

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/opal/service/proxy/ProxyInterface.java
```diff
@@ -5,7 +5,7 @@
 /** A marker interface to help identify code that is to be removed at a future date. */
 public interface ProxyInterface {
 
-    public static final String LEGACY_MODE = "LEGACY";
+    String LEGACY_MODE = "LEGACY";
```
RustyHMCTS, line 8: Why change this from static final? As a static final, it allows the compiler to optimise.

sabahirfan: Interface variables are static and final by default in Java.

RustyHMCTS (approved):

PR #210, opened 2024-02-20
PoC to demonstrate inserting files (as bytea) with Flyway

danlysiak:
### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/db/migration/CommonDbUtils.java
+++ b/src/main/java/db/migration/CommonDbUtils.java
@@ -1,0 +1,23 @@
+package db.migration;
+
+import org.flywaydb.core.api.migration.Context;
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class CommonDbUtils {
+
+    public static void insertFile(Context context, String name, String file) throws IOException {
+        JdbcTemplate jdbcTemplate = new JdbcTemplate(context.getConfiguration().getDataSource());
+        jdbcTemplate.update("INSERT INTO file_poc (name, value) VALUES ((?), (?))", name, readFileFromResources(file));
+    }
+
+    private static byte[] readFileFromResources(String filePath) throws IOException {
+        ClassPathResource resource = new ClassPathResource(filePath);
+        Path path = resource.getFile().toPath();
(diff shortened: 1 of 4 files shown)


src/main/java/db/migration/CommonDbUtils.java
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class CommonDbUtils {
+
+    public static void insertFile(Context context, String name, String file) throws IOException {
```
danlysiak, line 13: Depending on needs of table that we're actually inserting to, name this method more appropriately and pass in relevant args

src/main/java/db/migration/CommonDbUtils.java
```diff
@@ -0,0 +1,23 @@
...
+        JdbcTemplate jdbcTemplate = new JdbcTemplate(context.getConfiguration().getDataSource());
+        jdbcTemplate.update("INSERT INTO file_poc (name, value) VALUES ((?), (?))", name, readFileFromResources(file));
```
danlysiak, line 15: Modify this generic query appropriately

src/main/java/db/migration/V20240220_091__InsertTestFile.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.io.IOException;
+
+import static db.migration.CommonDbUtils.insertFile;
+
+public class V20240220_091__InsertTestFile extends BaseJavaMigration {
+
+    @Override
+    public void migrate(Context context) throws IOException {
+        insertFile(context, "someFileName", "db/files/sample.pdf");
```
danlysiak, line 14: We can add as many files as we want for a single migration

src/main/java/db/migration/CommonDbUtils.java
danlysiak, line 13: Depending on needs of table that we're actually inserting to, name this method more appropriately and pass in relevant args

src/main/java/db/migration/CommonDbUtils.java
danlysiak, line 15: Modify this generic query appropriately; we may want to either make it either more generic a query or have separate methods for different table inserts if there are different scenarios

PR #242, opened 2024-03-07
Pass Business Unit Id to/from FE

RustyHMCTS:
This is the first step for PO-235, passing the business unit id to the FE.
After The FE has coded changes to return that business unit id to us as part of an 'add notes' call, we can then check that the user has a role associated with that business unit that also has the 'add note' permission,.

```
[ ] Yes
[*] No
```


src/functionalTest/resources/features/opalMode/PO-184-187-188-192_addNote_postedBy_postedByAAD.feature
```diff
@@ -9,7 +9,7 @@ Feature: tests for changes to postedBy and postedByAAD
     And the add notes response contains
       | associatedRecordId | 500000010                                |
       | noteText           | test postedBy and PostedByAAD Opal user1 |
-      | postedBy           | L070KG                                   |
+      | postedBy           | L017KG                                   |
```
RustyHMCTS, line 12: This has only changed because the roles are current stored in a 'Set', so return order is non-deterministic

DeclanClarkeCGI (approved):

PR #246, opened 2024-03-12
Ensure user has permission to 'addNote' to defendant account

RustyHMCTS:
Guard against unauthorised users adding a note to a defendant account.

```
[*] Yes
[ ] No
```

User now needs a role with the requisite permission in the correct business unit.

### DIFF
--- a/src/main/java/uk/gov/hmcts/opal/util/HttpUtil.java
+++ b/src/main/java/uk/gov/hmcts/opal/util/HttpUtil.java
@@ -41,4 +41,5 @@ public class HttpUtil {
             .map(tokenService::extractPreferredUsername)
             .orElse(null);
     }
+
 }
--- a/src/test/java/uk/gov/hmcts/opal/controllers/NoteControllerTest.java
+++ b/src/test/java/uk/gov/hmcts/opal/controllers/NoteControllerTest.java
@@ -8,13 +8,11 @@ import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
-import uk.gov.hmcts.opal.authorisation.model.UserState;
 import uk.gov.hmcts.opal.dto.NoteDto;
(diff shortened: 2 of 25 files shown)


src/main/java/uk/gov/hmcts/opal/util/HttpUtil.java
```diff
@@ -41,4 +45,23 @@ public static String extractPreferredUsername(HttpServletRequest request, Access
...
+        }
+        return true;
+    }
+
+    public static boolean checkAnyRoleHasPermission(UserState userState, Permissions permission) {
+        if (userState.noRoleHasPermission(permission)) {
+            throw new AccessDeniedException("User does not have the required permission: " + permission.description);
+        }
+        return true;
+    }
```
sabahirfan, line 66: These methods can be moved in it's own utility class (i.e. RoleUtils or PermissionUtil etc.) rather than be in HttpUtil.

src/test/java/uk/gov/hmcts/opal/controllers/NoteControllerTest.java
```diff
@@ -123,4 +124,24 @@ public void testNotesSearch_NoContent() {
         verify(noteService, times(1)).searchNotes(any(
             NoteSearchDto.class));
     }
+
+    private UserState createUserState() {
```
sabahirfan, line 128: As this is duplicate builder method, can be moved to a helper class and reuse. i.e. 

```
class UserStateBuilder {

public static UserState createUserState() { ... }

}
```

src/main/java/uk/gov/hmcts/opal/util/HttpUtil.java
RustyHMCTS: Moved to a PermissionUtil class

src/test/java/uk/gov/hmcts/opal/controllers/NoteControllerTest.java
RustyHMCTS: Moved to a UserStateBuilder class

PR #254, opened 2024-03-13
PO-237 added script to change user_id from varchar to bigint and also…

abrahamdennis:
PO-237

### Change description ###
PO-237 added script to change user_id from varchar to bigint and also made corresponding changes where it is used as foreign key

```
[ ] Yes
[x] No
```

### DIFF
--- a/Sample_text.txt
+++ b/Sample_text.txt
@@ -1,0 +1,3 @@
+My name is Abraham Dennis
+Arsenal
+London!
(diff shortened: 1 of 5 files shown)


Sample_text.txt
```diff
@@ -0,0 +1,3 @@
+My name is Abraham Dennis
+Arsenal
```
RustyHMCTS, line 2: Hmmmm

RustyHMCTS (approved):

PR #300, opened 2024-04-03
PO-255 Add integration test for DefendantAccountController

RustyHMCTS:

```
[ ] Yes
[*] No
```


sabahirfan (changes requested):
Can we please add tests for all the endpoints/methods in this controller. These seems to be missing: `putDefendantAccount`  , `addNote`, `getNotesForDefendantAccount` etc.

sabahirfan (approved):

PR #317, opened 2024-04-17
PO-265: Add annotation AuthorizedPermission 

sabahirfan:
PO-265: 

### Change description ###
Add annotation AuthorizedPermission 

```
[x] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/opal/service/opal/NoteService.java
```diff
@@ -33,6 +35,7 @@ public class NoteService implements NoteServiceInterface {
 
     @Override
     @FeatureToggle(feature = "add-note", value = true)
+    @AuthorizedPermission(Permissions.ACCOUNT_ENQUIRY_NOTES)
```
RustyHMCTS, line 38: I'm not 100% on what you're doing with these annotations, but this looks wrong to me.
The check needs to be against a specific role, not any role within the user state.

RustyHMCTS (changes requested):

src/main/java/uk/gov/hmcts/opal/service/opal/NoteService.java
sabahirfan: This is doing exactly the same check we had previously in our controllers code. 
https://github.com/hmcts/opal-fines-service/pull/317/files#diff-7fd3e88c8ec744c457c69d92339fe51bd7bd295b7b4f2ad9dbb85ca49e5c4ae0R30-R33 

Previously here

RustyHMCTS: This is the check for the 'addNote/saveNote' - https://github.com/hmcts/opal-fines-service/blob/22731cf62149435abce80604ccec4e3f5e26c3df/src/main/java/uk/gov/hmcts/opal/controllers/DefendantAccountController.java#L134-L136

RustyHMCTS (changes requested):

src/main/java/uk/gov/hmcts/opal/service/opal/NoteService.java
sabahirfan: Oh yes, it's not needed here, this is different scenario though. removed.

src/main/java/uk/gov/hmcts/opal/controllers/LoggingExceptionResolver.java
```diff
@@ -1,33 +0,0 @@
...
-import org.springframework.web.servlet.ModelAndView;
-import uk.gov.hmcts.opal.authentication.service.AccessTokenService;
-
-import static uk.gov.hmcts.opal.authentication.service.AccessTokenService.AUTH_HEADER;
-import static uk.gov.hmcts.opal.util.HttpUtil.extractPreferredUsername;
-
-@Component
-@RequiredArgsConstructor
-@Slf4j(topic = "LoggingExceptionResolver")
-public class LoggingExceptionResolver implements HandlerExceptionResolver {
```
RustyHMCTS, line 19: I think we might still need this class to log the AccessDeniedException message from addNote/SaveNote not being permitted

sabahirfan: AccessDeniedException is being handeled in GlobalExceptionHandler now as we were doing other exceptions

RustyHMCTS (approved):

rajanigandra (approved):

PR #702, opened 2025-01-13
New test-support endpoint to call legacy gob test function

DeclanClarkeCGI:
New test support endpoint for legacy e2e sanity

```
[ ] Yes
[ X] No
```


src/integrationTest/java/uk/gov/hmcts/opal/controllers/TestingSupportControllerTest.java
```diff
@@ -66,6 +69,9 @@ class TestingSupportControllerTest {
     @MockBean
     private AuthorisationService authorisationService;
 
+    @MockBean
```
RustyHMCTS, line 72: I think @MockBean is either deprecated, or soon to be. The replacement should be @MockitoBean (from memory). Doesn't need to be changed now, just notifying for reference,

RustyHMCTS (approved):

src/integrationTest/java/uk/gov/hmcts/opal/controllers/TestingSupportControllerTest.java
DeclanClarkeCGI: Good to know, will bear that in mind going forward

PR #1171, opened 2025-09-17
PO-1566-Add-Note

marcmollins:

https://tools.hmcts.net/jira/browse/PO-1566

### Change description ###

adding an API for notes within OPAL

```
[ ] Yes
[x] No
```


RustyHMCTS (changes requested):
I think we need to wait for PO-2117 (ETag) to be merged so that we can integrate those methods.

PR #1223, opened 2025-10-07
PO-2222_PO-2244 - Create frontend test data and update DEFENDANT_ACCOUNT.ACCOUNT_STATUS description

Chi-cgi:
JIRA link (if applicable)

https://tools.hmcts.net/jira/browse/PO-2222

https://tools.hmcts.net/jira/browse/PO-2244

Change description

PO-2222 - Created database test data to support frontend development testing.
PO-2244 - Updated the description of DEFENDANT_ACCOUNT.ACCOUNT_STATUS to align with the Data Model spreadsheet.

Does this PR introduce a breaking change?** (check one with "x")**

[ ] Yes
[x] No


iamfrankiemoran:
Initial look at the test data looks good. Couple of observations:

1. `account_numbers` in the `defendant_accounts` table are not correct. The format is 8 digits with an optional check character at the end. Examples: 12345678 or 12345678A. As this would result in an issue for us testing the account number search: 
image
2. `prosecutor_case_reference` in the `defendant_accounts` table are not correct. The format should only accept numbers and letters. This would result in an issue for us testing the `prosecutor_case_reference` via the frontend: 
image
3. I can see Individual accounts which is great, could we also get some Companies added too? I can also see Minor Creditor `Minor Creditor Test Ltd`. However, when we search for this using the Search for an account screen it shows 10 times: 
image
Could I just double check this is correct?
4. The parties which have `TestSurname6` and `TestForename6` etc. can't have numbers in them as `surname` and `forenames` must only contain letters. There is frontend validation enforced therefore we couldn't search for these accounts unless we search for `TestSurname` which brings back all 5 accounts. 
image
5. Could we get some data for Fixed Penalty please? 
6. Could we get a couple of accounts where it is a Parent or Guardian To Pay please? 
7. We have old business unit data in the `business_units` table (i.e. 42 business units where `opal_domain` is null which I believe are duplicated by their newer version where `opal_domain` are populated)
8. Could we also look at getting the `draft_accounts` table populated with this please? https://tools.hmcts.net/confluence/display/PO/Draft+Accounts

PR #1327, opened 2025-11-07
PO-1589-replace-defendant-account-party

marcmollins:

https://tools.hmcts.net/jira/browse/PO-1589

### Change description ###

changes for replacing a defendant account party

```
[ ] Yes
[x] No
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: actions/setup-java@v5
         with:
           distribution: 'temurin' # See 'Supported distributions' for available options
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountService.java
```diff
@@ -1046,37 +1117,387 @@ public DefendantAccountResponse updateDefendantAccount(
             .build();
     }
 
-    private static String orEmpty(String s) {
-        return (s == null) ? "" : s;
+    private DefendantAccountEntity bumpVersion(Long accountId) {
+        DefendantAccountEntity entity = getDefendantAccountById(accountId);
+        entity.setVersion(entity.getVersion() + 1);
+        return defendantAccountRepository.saveAndFlush(entity);
+    }
```
DeclanClarkeCGI: I don't think this is safe way to update the version. You aren't comparing the version received and the current db version atomically.

Could use em.lock(entity, LockModeType.OPTIMISTIC_FORCE_INCREMENT); in your transactional method

PR #1385, opened 2025-12-08
Po 1918 add defendant account enforcement legacy

marcmollins:
https://tools.hmcts.net/jira/browse/PO-1918

### Change description ###

legacy mode for adding enforcement to defendant account

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -85,7 +85,7 @@ configurations {
   smokeTestRuntimeOnly.extendsFrom runtimeOnly
 }
 
-configurations.all {
+configurations.configureEach {
   resolutionStrategy {
     force "io.cucumber:cucumber-core:${cucumberVersion}"
     force "io.cucumber:cucumber-java:${cucumberVersion}"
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/opal/controllers/DefendantAccountController.java
```diff
@@ -183,4 +180,23 @@ public ResponseEntity<GetDefendantAccountPartyResponse> replaceDefendantAccountP
                 defendantAccountPartyId, authHeaderValue, ifMatch, businessUnitId, request));
     }
 
+    @PostMapping("/{defendantAccountId}/enforcements")
+    @Operation(summary = "Create an enforcement for a given defendant account")
+    public ResponseEntity<AddEnforcementResponse> addDefendantAccountEnforcement(
```
ShanCGI: Can we rename this to addEnforcement() please?

That'll make it consistent with the naming in the interfaces and services. Also, since the navigation starts out from Defendant Account (and this is the DefendantAccountController), the obvious prefix can be skipped to keep the method names shorter.

PR #1628, opened 2026-03-26
PO-1854 update defendant account clean open api implementation

HennaParmar:
https://tools.hmcts.net/jira/browse/PO-1854

Change description
Updated the existing PATCH /defendant-accounts/{id} endpoint to support amendment of defendant enforcement attributes, including collection order fields, enforcing court, and enforcement override values, in line with the TDIA and API specification. The change also covers validation, permission enforcement for Account Maintenance, legacy-mode support, amendment auditing, and integration test coverage for happy path and error scenarios.

Does this PR introduce a breaking change? (check one with "x")

[ ] Yes
[ x ] No


sarahtw-cgi (commented):
Just a couple of comments/questions

PR #1697, opened 2026-04-16
Po 1775 remove defendant account enforcement hold opal mode]

HennaParmar:
https://tools.hmcts.net/jira/browse/PO-1775

### Change description ###
New functional endpoint to remove an Enforcement Hold for a given Defendant Account, and creation / update of associated records as necessary 

```
[ ] Yes
[ x ] No
```


src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
```diff
@@ -56,14 +74,96 @@ public AddEnforcementResponse addEnforcement(
...
+
+        if (ifMatch == null || ifMatch.isBlank()) {
+            throw new ResourceConflictException(
+                "Defendant Account",
+                defendantAccountId,
+                "If-Match header is required",
+                defendantEntity
+            );
+        }
+
```
marcmollins, line 100: don't think this is needed, if there is an issue with the version, the next method won't extract it and an error will be thrown.

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
```diff
@@ -56,14 +74,96 @@ public AddEnforcementResponse addEnforcement(
...
+                ifMatch,
+                userState,
+                savedEntity
+            );
+
+            reportEntryService.createRemoveEnforcementHoldReportEntry(defendantAccountId, businessUnitId);
+
+            amendmentService.auditFinaliseStoredProc(
+                defendantAccountId,
+                uk.gov.hmcts.opal.entity.amendment.RecordType.DEFENDANT_ACCOUNTS,
```
marcmollins, line 134: there is 2 record types that are very similar, might be worth consolidating them, means you don't have to have full qualified name here and cleans up debt.

marcmollins (changes requested):

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
HennaParmar: good spot - I've taken it out now

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
HennaParmar: added a small consolidation here so the enforcement flow can use the existing DTO RecordType directly and AmendmentService maps it internally for the audit stored proc calls. That removes the fully qualified enum usage at this call site without widening the change beyond this path.

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
HennaParmar: Taking it out broke the pipeline so I’ve put it back in now. Without it, a blank header falls through to VersionUtils and we get a NumberFormatException rather than the intended ResourceConflictException with the clear "If-Match header is required" response, which also broke the existing test. This keeps the API behaviour explicit for missing headers and still leaves version parsing/validation to the next method for non-blank values.

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
HennaParmar: Reverted it back to the original version now as discussed 👍

marcmollins (approved):

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
```diff
@@ -56,14 +74,96 @@ public AddEnforcementResponse addEnforcement(
...
+                businessUnitUserId,
+                null,
+                "Remove Enforcement Hold"
+            );
+
+            return RemoveDefendantAccountEnforcementHoldResponse.builder()
+                .defendantAccountId(String.valueOf(savedEntity.getDefendantAccountId()))
+                .version(savedEntity.getVersion())
+                .build();
+
```
Sean-Reed-CGI, line 141: We usually let ObjectOptimisticLockingFailureException bubble up and be handled by the GlobalExceptionHandler. Is there a reason why we aren't just doing that here?

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
```diff
@@ -56,14 +74,96 @@ public AddEnforcementResponse addEnforcement(
...
+                "Account version has changed",
+                defendantEntity
+            );
+        }
+    }
+
+    private AddNoteRequest buildRemoveEnforcementHoldNoteRequest(
+        Long defendantAccountId,
+        RemoveDefendantAccountEnforcementHoldRequest request) {
+
```
Sean-Reed-CGI, line 151: Might be nicer if Note had a builder method

src/main/java/uk/gov/hmcts/opal/service/opal/ReportEntryService.java
```diff
@@ -43,4 +45,19 @@ public void createExtendTtpReportEntry(Long paymentTermsId, short businessUnitId
         log.debug(":createExtendTtpReportEntry: created report entry for paymentTermsId={} BU={}",
             paymentTermsId, businessUnitId);
     }
+
+    public void createRemoveEnforcementHoldReportEntry(Long defendantAccountId, short businessUnitId) {
+        ReportEntryEntity reportEntry = new ReportEntryEntity();
```
Sean-Reed-CGI: Can we use the builder method to construct the ReportEntryEntity

src/test/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementServiceTest.java
```diff
@@ -0,0 +1,319 @@
...
+            .defendantAccountId(defendantAccountId)
+            .lastEnforcement("NOENF")
+            .versionNumber(7L)
+            .build();
+
+        UserState userState = allPermissionsUser();
+
+        when(userStateService.checkForAuthorisedUser(authHeader)).thenReturn(userState);
+        when(defendantAccountRepositoryService.findById(defendantAccountId)).thenReturn(defendantEntity);
+        when(defendantAccountRepositoryService.saveAndFlush(any(DefendantAccountEntity.class)))
```
Sean-Reed-CGI: Avoid using 'any' Mockito methods where possible. Here we need to verify that the defendantAccount passed to saveAndFlush is the one returned from the repository

And can you add in //arrange //act //assert comments to make the test more readable

src/main/java/uk/gov/hmcts/opal/service/opal/ReportEntryService.java
```diff
@@ -43,4 +45,19 @@ public void createExtendTtpReportEntry(Long paymentTermsId, short businessUnitId
...
+
+        reportEntry.setBusinessUnit(businessUnitService.getBusinessUnit(businessUnitId));
+        reportEntry.setReportId(TRACK_ENFORCEMENT_HOLD);
+        reportEntry.setEntryTimestamp(LocalDateTime.now());
```
Sean-Reed-CGI: It's best to use LocalDateTime.now(clock) where the Clock is a wired in spring bean.
This will make testing easier.

src/test/java/uk/gov/hmcts/opal/service/DefendantAccountEnforcementServiceTest.java
```diff
@@ -289,4 +291,181 @@ void removeEnforcementHold_whenUserLacksPermission_throwsPermissionNotAllowedExc
         verify(userState).hasBusinessUnitUserWithPermission((short) 10, FinesPermission.ENTER_ENFORCEMENT);
         verifyNoInteractions(defendantAccountEnforcementServiceProxy);
     }
+
```
Sean-Reed-CGI, line 300: The unit under test here didn't change in this PR, so is this just filling in a missing unit test you spotted?
Would be more readable with //arrange //act //assert

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
```diff
@@ -56,14 +74,96 @@ public AddEnforcementResponse addEnforcement(
...
+        }
+
+        amendmentService.auditInitialiseStoredProc(
+            defendantAccountId,
+            uk.gov.hmcts.opal.entity.amendment.RecordType.DEFENDANT_ACCOUNTS
+        );
+
+        try {
+            defendantEntity.setLastEnforcement(null);
+            defendantEntity.setLastMovementDate(LocalDate.now());
```
Sean-Reed-CGI: Use a wired in Clock to get the current time

Sean-Reed-CGI (changes requested):
Added a few comments

src/test/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementServiceTest.java
```diff
@@ -0,0 +1,319 @@
...
+                    ifMatch,
+                    authHeader,
+                    request
+                );
+
+            assertNotNull(result);
+            assertEquals(String.valueOf(defendantAccountId), result.getDefendantAccountId());
+            assertEquals(BigInteger.valueOf(7L), result.getVersion());
+            assertNull(defendantEntity.getLastEnforcement());
+            assertNotNull(defendantEntity.getLastMovementDate());
```
Sean-Reed-CGI: We should assert that the current time is set here

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
HennaParmar: updated to use the wired Clock bean instead of LocalDate.now() directly, and adjusted the unit test to use a fixed clock.

src/test/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementServiceTest.java
HennaParmar: Thanks - updated this test to stub saveAndFlush with the exact DefendantAccountEntity returned from the repository rather than using any(...), and added arrange/act/assert comments for readability.

src/main/java/uk/gov/hmcts/opal/service/opal/ReportEntryService.java
HennaParmar: Updated this to use the existing wired Clock bean for the entry timestamp and switched the ReportEntryEntity construction over to the builder. I also tightened the unit test to use a fixed clock and assert the exact timestamp.

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
HennaParmar: thanks - updated this to follow the usual repo pattern. I removed the local catch so ObjectOptimisticLockingFailureException now bubbles up to the existing GlobalExceptionHandler, and I updated the unit test to assert the original exception is propagated.

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
HennaParmar: Thanks - added a builder to the Note DTO and switching the remove-enforcement-hold note construction to use Note.builder() instead of manual setters.

src/test/java/uk/gov/hmcts/opal/service/DefendantAccountEnforcementServiceTest.java
HennaParmar: This was an opportunistic missing test I added while working in the same area. I’ve now updated it to include arrange/act/assert comments to make the flow clearer.

src/test/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementServiceTest.java
HennaParmar: updated this assertion to check the exact date from the fixed clock rather than just asserting non-null, so the current time being set is now covered explicitly.

src/main/java/uk/gov/hmcts/opal/service/opal/ReportEntryService.java
HennaParmar: Updated this to construct the report entry with ReportEntryEntity.builder() rather than using new ReportEntryEntity().

src/test/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementServiceTest.java
Sean-Reed-CGI: You fixed the issues I mentioned but there is more use of any() in that unit test that need improving.

HennaParmar: Thanks - Updated the remaining broad Mockito matchers in this test as well. I removed the unnecessary any(AddNoteRequest.class) stub and replaced the auditFinaliseStoredProc never-verification to use the exact expected arguments rather than generic any() matchers.

Sean-Reed-CGI (approved):
LGTM

jonathanDuffy (approved):

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
```diff
@@ -56,14 +76,87 @@ public AddEnforcementResponse addEnforcement(
...
+                "Defendant Account",
+                defendantAccountId,
+                "No enforcement hold to remove",
+                defendantEntity
+            );
+        }
+
+        amendmentService.auditInitialiseStoredProc(
+            defendantAccountId,
+            uk.gov.hmcts.opal.entity.amendment.RecordType.DEFENDANT_ACCOUNTS
```
Ben-Edwards-cgi: Should import RecordType instead of fully quantifying it

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
```diff
@@ -56,14 +76,87 @@ public AddEnforcementResponse addEnforcement(
...
+            VersionUtils.createETag(savedEntity),
+            userState,
+            savedEntity
+        );
+
+        reportEntryService.createRemoveEnforcementHoldReportEntry(defendantAccountId, businessUnitId);
+
+        amendmentService.auditFinaliseStoredProc(
+            defendantAccountId,
+            uk.gov.hmcts.opal.entity.amendment.RecordType.DEFENDANT_ACCOUNTS,
```
Ben-Edwards-cgi: Should import RecordType instead of fully quantifying it

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
HennaParmar: I’ve removed the duplicate uk.gov.hmcts.opal.entity.amendment.RecordType enum and switched the amendment audit calls over to use uk.gov.hmcts.opal.dto.RecordType instead. I also updated the service/repository methods that were taking the old enum so they now use the DTO enum and pass its string value through to the stored proc calls.

src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountEnforcementService.java
HennaParmar: I’ve removed the duplicate uk.gov.hmcts.opal.entity.amendment.RecordType enum and switched the amendment audit calls over to use uk.gov.hmcts.opal.dto.RecordType instead. I also updated the service/repository methods that were taking the old enum so they now use the DTO enum and pass its string value through to the stored proc calls.

Ben-Edwards-cgi (approved):

PR #1734, opened 2026-04-28
PO-1991 - Get Minor Creditor Account Legacy

Aneesah33:
https://tools.hmcts.net/jira/browse/PO-1991

### Change description ###

- Adds legacy-mode support for GET /minor-creditor-accounts/{id}
- Wires the endpoint through controller, shared service, proxy, and legacy service
- Uses legacy gateway actionType GET_MINOR_CREDITOR_ACCOUNT_PARTY with accountId request mapping
- Maps legacy response DTOs into the API response using legacy-specific mappers
- Applies Search and View Accounts auth and filters BACS details unless the user has View Creditor BACS in the account business unit
- Leaves Opal mode unsupported in this ticket
- Adds test coverage for controller/service/legacy service and legacy integration scenarios
- Adds matching legacy stub mapping and XML response in opal-legacy-db-stub

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/opal/controllers/MinorCreditorApiController.java
+++ b/src/main/java/uk/gov/hmcts/opal/controllers/MinorCreditorApiController.java
@@ -1,6 +1,7 @@
 package uk.gov.hmcts.opal.controllers;
 
 import static uk.gov.hmcts.opal.util.HttpUtil.buildCreatedResponse;
+import static uk.gov.hmcts.opal.util.HttpUtil.buildResponse;
 import static uk.gov.hmcts.opal.util.VersionUtils.extractOptionalBigInteger;
 
 import java.util.Optional;
@@ -28,6 +29,15 @@ public class MinorCreditorApiController implements MinorCreditorApi {
     @Override
     public Optional<NativeWebRequest> getRequest() {
         return Optional.ofNullable(request);
+    }
+
+    @Override
+    public ResponseEntity<MinorCreditorAccountResponseMinorCreditor> getMinorCreditorAccount(Long id) {
+        log.debug(":GET:getMinorCreditorAccount: id={}", id);
+
+        MinorCreditorAccountResponse result = minorCreditorService.getMinorCreditorAccount(id);
+
+        return buildResponse(result);
     }
 
     @Override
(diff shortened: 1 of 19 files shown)


src/main/java/uk/gov/hmcts/opal/controllers/MinorCreditorApiController.java
```diff
@@ -27,6 +27,18 @@ public Optional<NativeWebRequest> getRequest() {
         return Optional.ofNullable(request);
     }
```
Sean-Reed-CGI, line 25: Looks like the use of NativeWebRequest stored as a field in this controller wouldn't be thread-safe

Aneesah33: Agreed, I spoke to Ben and we aligned on using Spring Security context for this path. I’ve removed the NativeWebRequest handling from the controller, so it no longer stores request-specific state, and the authenticated user is now resolved via UserStateService from Spring Security context.

DeclanClarkeCGI (commented):
The API has been designed as Get Minor Creditor Account but we have some references to Get Minor Creditor Account Party, here and in the legacy stub. I appreciate there are party details in the response but it would be good if we can be consistent.

DeclanClarkeCGI (approved):

Sean-Reed-CGI:
Please make sure all new unit tests have the // Arrange // Act // Assert comments

src/test/java/uk/gov/hmcts/opal/service/UserStateServiceTest.java
```diff
@@ -52,6 +55,24 @@ void testCheckForAuthorisedUser_success() {
         assertEquals(1L, userState.getUserId());
     }
```
Sean-Reed-CGI: Missing unit tests for authorization==null and when principal is not a Jwt

HennaParmar: Added both missing cases in `UserStateServiceTest`:
- `authorization == null`
- security context principal is not a `Jwt`

This now covers the new `SecurityContext` path and the null-authorization fallback explicitly.

Sean-Reed-CGI (approved):
LGTM

CadeFaulkner (approved):

PR #1750, opened 2026-05-04
Po 1992 update minor creditor

HennaParmar:
https://tools.hmcts.net/jira/browse/PO-1992

PO-1992 - BE Update Minor Creditor [Opal mode]
Implemented and verified the Opal-mode PATCH /minor-creditor-accounts/{id} update path for minor creditor account amendments. Adds Account Details view for defendant accounts from search, with header summary, “At a Glance” tab, note management, and Welsh language preference support.

```
[ ] Yes
[ x ] No
```


src/integrationTest/java/uk/gov/hmcts/opal/controllers/MinorCreditorApiControllerFeatureFlagIntegrationTest.java
```diff
@@ -0,0 +1,154 @@
...
+                              .organisationFlag(false)
+                              .individualDetails(new IndividualDetailsCommon()
+                                                     .forenames("Creditor")
+                                                     .surname("Updated")))
+            .address(new AddressDetailsCommon()
+                         .addressLine1("99 Updated Road")
+                         .postcode("NW1 1AA"))
+            .payment(new CreditorAccountPaymentDetailsCommon()
+                         .holdPayment(true)
+                         .payByBacs(true));
```
Aneesah33, line 85: We send payByBacs(true) here, but the test doesn’t check that it was actually saved. Should we assert this too, so we know the payment update works?

src/integrationTest/java/uk/gov/hmcts/opal/controllers/MinorCreditorControllerIntegrationTest.java
```diff
@@ -410,49 +592,39 @@ void patchMinorCreditor_missingPayload_returns400() throws Exception {
...
+            .partyDetails(new PartyDetailsCommon()
+                              .partyId("99007")
+                              .organisationFlag(false)
+                              .individualDetails(new IndividualDetailsCommon()
+                                                     .surname("Deleted")))
+            .address(new AddressDetailsCommon()
+                         .addressLine1("33 Delete St.")
+                         .postcode("DE1 2DE"))
+            .payment(new CreditorAccountPaymentDetailsCommon()
+                         .payByBacs(true)
```
Aneesah33, line 630: Can we also test the bank payment fields from the tdia, like account name, sort code, account number and payment reference? Right now this only checks holdPayment.

src/main/java/uk/gov/hmcts/opal/service/opal/OpalMinorCreditorService.java
```diff
@@ -143,6 +143,7 @@ public MinorCreditorAccountResponse updateMinorCreditorAccount(
         updateMapper.updateParty(request.getPartyDetails(), request.getAddress(), party);
 
         creditorAccount.setHoldPayout(request.getPayment().getHoldPayment());
```
Aneesah33, line 150: Should we save the other payment fields here too, like payByBacs, account name, sort code, account number and account reference? At the moment only holdPayment is updated.

src/integrationTest/java/uk/gov/hmcts/opal/controllers/MinorCreditorApiControllerFeatureFlagIntegrationTest.java
```diff
@@ -0,0 +1,154 @@
...
+import uk.gov.hmcts.opal.generated.model.PatchMinorCreditorAccountRequest;
+import uk.gov.hmcts.opal.generated.model.PartyDetailsCommon;
+import uk.gov.hmcts.opal.repository.CreditorAccountRepository;
+import uk.gov.hmcts.opal.service.UserStateService;
+import uk.gov.hmcts.opal.util.FeatureFlags;
+
+@ActiveProfiles({"integration", "opal"})
+@TestPropertySource(properties = {
+    "launchdarkly.enabled=true"
+})
```
DeclanClarkeCGI, line 34: I think we said we would make this false and just test the default property behaviour in the fines service

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalMinorCreditorIntegrationTest.java
```diff
@@ -33,25 +39,35 @@
 import uk.gov.hmcts.opal.repository.PartyRepository;
 import uk.gov.hmcts.opal.repository.jpa.CreditorTransactionSpecs;
 import uk.gov.hmcts.opal.repository.jpa.ImpositionSpecs;
+import uk.gov.hmcts.opal.util.FeatureFlags;
 
 @ActiveProfiles({"integration", "opal"})
-@Sql(scripts = "classpath:db/insertData/insert_into_minor_creditors.sql", executionPhase = BEFORE_TEST_CLASS)
-@Sql(scripts = "classpath:db/deleteData/delete_from_minor_creditors.sql", executionPhase = AFTER_TEST_CLASS)
+@TestPropertySource(properties = "launchdarkly.enabled=true")
```
DeclanClarkeCGI: I think we said we would make this false and just test the default property behaviour in the fines service

src/main/java/uk/gov/hmcts/opal/service/MinorCreditorService.java
```diff
@@ -104,6 +105,13 @@ public MinorCreditorAccountResponse updateMinorCreditorAccount(
                 businessUnitIdShort,
                 FinesPermission.ACCOUNT_MAINTENANCE);
         }
+        if (isUpdatingPayoutHold(minorCreditorId, businessUnitIdShort, request)
+            && !userState.hasBusinessUnitUserWithPermission(businessUnitIdShort,
```
Aneesah33: TL clarified that ADD_AND_REMOVE_PAYMENT_HOLD should apply to the entire payment object, since it includes hold_payment. Should we update this logic to check the permission whenever a payment object is present, rather than comparing the current value?

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalMinorCreditorIntegrationTest.java
```diff
@@ -33,25 +39,35 @@
...
 
-    @MockitoSpyBean
+    @Autowired
     private CreditorAccountRepository creditorAccountRepository;
 
+    @BeforeEach
+    void setUpFeatureFlag() {
+        when(featureToggleApi.isFeatureEnabled(eq(FeatureFlags.RELEASE_1B), anyBoolean())).thenReturn(true);
+    }
+
```
DeclanClarkeCGI: If we test with LD disabled then we shouldn't need to stub

build.gradle
```diff
@@ -567,6 +567,7 @@ dependencies {
   }
   testImplementation 'org.junit.platform:junit-platform-suite:6.0.3'
   testImplementation "io.cucumber:cucumber-junit-platform-engine:${cucumberVersion}"
+  integrationTestImplementation 'com.launchdarkly:launchdarkly-java-server-sdk:7.13.4'
```
DeclanClarkeCGI: Do we need this, all LD integration should be contained to the common-lib ?

src/integrationTest/java/uk/gov/hmcts/opal/controllers/MinorCreditorApiControllerFeatureFlagIntegrationTest.java
HennaParmar: Removed this LD-specific integration test. The common-lib repo will carry the LaunchDarkly integration coverage, and in fines-service I’ve kept the fallback-property path instead of testing LD-enabled behaviour here.

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalMinorCreditorIntegrationTest.java
HennaParmar: Changed this to run with launchdarkly.enabled=false and launchdarkly.default-flag-values.release-1b=true, so the fines-service tests now exercise the local default property behaviour rather than LD-enabled behaviour.

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalMinorCreditorIntegrationTest.java
HennaParmar: Removed the FeatureToggleApi stubbing. With LD disabled for this test class, the endpoint now relies on the fallback property instead, so the extra mock setup was redundant.

build.gradle
HennaParmar: Removed this dependency. With the LD-specific integration coverage dropped from fines-service, the extra LaunchDarkly SDK test classpath entry is no longer needed here.

src/main/java/uk/gov/hmcts/opal/service/MinorCreditorService.java
HennaParmar: Cheers - updated this to match the TL clarification. We now require ADD_AND_REMOVE_PAYMENT_HOLD whenever a payment object is supplied, rather than comparing the current hold_payment value. I also aligned the service and integration tests to the new 403 behaviour.

Aneesah33 (approved):
LGTM

DeclanClarkeCGI (approved):

vibhu-HMCTS (approved):

PR #1752, opened 2026-05-05
PO-1993 update minor creditor (legacy)

HennaParmar:
https://tools.hmcts.net/jira/browse/PO-1993

### Change description ###
PO-1993 implement legacy minor creditor PATCH update flow

- add legacy request/response DTOs for minor creditor PATCH
- map PATCH payload to legacy gateway request and response back to opal DTO
- implement legacy update path in LegacyMinorCreditorService
- normalise legacy 4xx gateway errors via GlobalExceptionHandler
- add legacy service and integration coverage for PATCH happy/unhappy paths

```
[ ] Yes
[ x ] No
```


src/main/java/uk/gov/hmcts/opal/mapper/legacy/LegacyUpdateMinorCreditorAccountResponseMapper.java
```diff
@@ -0,0 +1,124 @@
...
+import uk.gov.hmcts.opal.dto.legacy.common.OrganisationDetails;
+import uk.gov.hmcts.opal.generated.model.AddressDetailsCommon;
+import uk.gov.hmcts.opal.generated.model.IndividualAliasCommon;
+import uk.gov.hmcts.opal.generated.model.IndividualDetailsCommon;
+import uk.gov.hmcts.opal.generated.model.MinorCreditorAccountResponseMinorCreditorPayment;
+import uk.gov.hmcts.opal.generated.model.OrganisationAliasCommon;
+import uk.gov.hmcts.opal.generated.model.OrganisationDetailsCommon;
+import uk.gov.hmcts.opal.generated.model.PartyDetailsCommon;
+
+@Component
```
sarahtw-cgi: Mapstruct could be used here

src/main/java/uk/gov/hmcts/opal/mapper/request/UpdateMinorCreditorAccountRequestMapper.java
```diff
@@ -0,0 +1,127 @@
...
+import uk.gov.hmcts.opal.dto.legacy.common.OrganisationDetails;
+import uk.gov.hmcts.opal.generated.model.AddressDetailsCommon;
+import uk.gov.hmcts.opal.generated.model.CreditorAccountPaymentDetailsCommon;
+import uk.gov.hmcts.opal.generated.model.IndividualAliasCommon;
+import uk.gov.hmcts.opal.generated.model.IndividualDetailsCommon;
+import uk.gov.hmcts.opal.generated.model.OrganisationAliasCommon;
+import uk.gov.hmcts.opal.generated.model.OrganisationDetailsCommon;
+import uk.gov.hmcts.opal.generated.model.PartyDetailsCommon;
+import uk.gov.hmcts.opal.generated.model.PatchMinorCreditorAccountRequest;
+
```
sarahtw-cgi, line 24: should be using mapstruct

src/main/java/uk/gov/hmcts/opal/service/MinorCreditorService.java
```diff
@@ -93,8 +91,7 @@ public MinorCreditorAccountResponse updateMinorCreditorAccount(
         Short businessUnitIdShort = businessUnitId != null ? Short.valueOf(businessUnitId) : null;
 
         UserState userState = userStateService.checkForAuthorisedUser(authHeaderValue);
-        if (businessUnitId == null || !userState.hasBusinessUnitUserWithPermission(businessUnitIdShort,
-            FinesPermission.ADD_AND_REMOVE_PAYMENT_HOLD)) {
+        if (businessUnitId == null) {
             throw new PermissionNotAllowedException(
                 businessUnitIdShort,
```
sarahtw-cgi: businessUnitIdShort is always null here

sarahtw-cgi (changes requested):

src/main/java/uk/gov/hmcts/opal/mapper/legacy/LegacyUpdateMinorCreditorAccountResponseMapper.java
HennaParmar: Thanks - Updated this to MapStruct. The manual response mapper has been replaced with a Spring MapStruct mapper, with only the explicit conversions left in place for version, alias sequence numbers, LocalDate -> String, and firstNames -> forenames.

src/main/java/uk/gov/hmcts/opal/mapper/request/UpdateMinorCreditorAccountRequestMapper.java
HennaParmar: Cheers - Updated this to use MapStruct in line with the mapper guidance. The mapper is now declarative and only keeps the explicit conversions needed for header/path values and the few field-name/type differences; nested/null handling is left to MapStruct.

src/main/java/uk/gov/hmcts/opal/service/MinorCreditorService.java
HennaParmar: Good spot. I’ve moved the Short.valueOf(...) conversion to after the null guard and changed that exception path to pass an explicit (Short) null instead. That keeps the behaviour the same but removes the misleading businessUnitIdShort use in the null branch.

sarahtw-cgi (approved):

Ju-Morgan (approved):
ALGTM

CadeFaulkner (approved):

src/main/java/uk/gov/hmcts/opal/mapper/legacy/LegacyUpdateMinorCreditorAccountResponseMapper.java
```diff
@@ -0,0 +1,76 @@
...
+    builder = @Builder(disableBuilder = true)
+)
+public interface LegacyUpdateMinorCreditorAccountResponseMapper {
+
+    @Mappings({
+        @Mapping(target = "version", source = "accountVersion", qualifiedByName = "integerToBigInteger"),
+        @Mapping(target = "creditorAccountId", source = "creditorAccountId"),
+        @Mapping(target = "partyDetails", source = "partyDetails"),
+        @Mapping(target = "address", source = "address"),
+        @Mapping(target = "payment", source = "payment")
```
Sean-Reed-CGI, line 38: If target and source are the same think MapStruct will do this automatically

src/integrationTest/java/uk/gov/hmcts/opal/controllers/LegacyMinorCreditorPatchIntegrationTest.java
```diff
@@ -0,0 +1,236 @@
...
+            eq(null)
+        )).thenReturn(new GatewayService.Response<>(HttpStatus.OK, legacyPatchResponse()));
+
+        ResultActions resultActions = performLegacyPatch(PATCH_MINOR_CREDITOR_ACCOUNT_ID, "\"1\"");
+
+        String body = resultActions.andReturn().getResponse().getContentAsString();
+        log.info(":patchMinorCreditor_success_returns201 body:\n{}",
+            ToJsonString.toPrettyJson(body));
+
+        resultActions.andExpect(status().isCreated())
```
Aneesah33: Should this be asserting 200 OK rather than 201 Created? Jira AC7 lists as OK (200), and this endpoint is a PATCH update rather than creation. Otherwise the controller behaviour and tests should probably be aligned to 200.

src/integrationTest/java/uk/gov/hmcts/opal/controllers/LegacyMinorCreditorPatchIntegrationTest.java
```diff
@@ -0,0 +1,236 @@
...
+@Slf4j(topic = "opal.LegacyMinorCreditorPatchIntegrationTest")
+class LegacyMinorCreditorPatchIntegrationTest extends MinorCreditorControllerIntegrationTest {
+
+    private static final String AUTH_HEADER = "Bearer some_value";
+    private static final String URL_BASE = "/minor-creditor-accounts";
+    private static final long PATCH_MINOR_CREDITOR_ACCOUNT_ID = 607L;
+    private static final short PATCH_MINOR_CREDITOR_BUSINESS_UNIT_ID = 10;
+    private static final String UPDATE_MINOR_CREDITOR_ACCOUNT = "LIBRA.of_update_minor_creditor_account";
+
+    @MockitoBean
```
Aneesah33, line 63: Just checking, since GatewayService is mocked here, this test won’t hit the WireMock mappings added in opal-legacy-db-stub. Do we need some stub-backed tests as well to confirm the actionType and request body match those mappings?

src/integrationTest/java/uk/gov/hmcts/opal/controllers/LegacyMinorCreditorPatchIntegrationTest.java
HennaParmar: Thanks good spot - AC9 defines success as 200 OK, and this PATCH is an update rather than a create. I’ve aligned the controller behavior and the related tests to 200.

src/integrationTest/java/uk/gov/hmcts/opal/controllers/LegacyMinorCreditorPatchIntegrationTest.java
HennaParmar: Thanks I’ve kept the mocked integration test for controller/service coverage, and added a separate stub-backed legacy PATCH integration test in opal-fines-service. The paired PO-1993 stub mappings now assert the legacy actionType plus key mapped request fields, so we can verify the real gateway request shape against opal-legacy-db-stub.”

src/main/java/uk/gov/hmcts/opal/dto/legacy/LegacyUpdateMinorCreditorAccountRequest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @JsonProperty("account_version")
+    @XmlElement(name = "account_version")
+    @NotNull
+    private Integer accountVersion;
+
+    @JsonProperty("party_details")
+    @XmlElement(name = "party_details")
+    @NotNull
+    private LegacyPartyDetails partyDetails;
```
Aneesah33, line 52: The legacy schema expects snake_case nested fields, e.g. party_details.party_id, party_details.organisation_flag, and address.address_line_1. In the local stub log the outbound request is using camelCase for these nested objects (partyId, organisationFlag, addressLine1), so the stub does not match. Could we add/verify JSON serialization for LegacyPartyDetails and AddressDetailsLegacy so the outbound legacy request matches the schema?

src/main/java/uk/gov/hmcts/opal/dto/legacy/LegacyUpdateMinorCreditorAccountRequest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @JsonProperty("address")
+    @XmlElement(name = "address")
+    @NotNull
+    private AddressDetailsLegacy address;
```
Aneesah33, line 57: Same as above for AddressDetailsLegacy.

src/test/java/uk/gov/hmcts/opal/dto/legacy/LegacyUpdateMinorCreditorAccountRequestTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+        JsonNode json = ToJsonString.getObjectMapper().readTree(
+            ToJsonString.getObjectMapper().writeValueAsString(request)
+        );
+
+        assertEquals("607", json.get("creditor_account_id").asText());
+        assertEquals("99008", json.get("party_details").get("party_id").asText());
+        assertTrue(json.get("party_details").get("organisation_flag").asBoolean());
+        assertEquals("Updated Ltd",
+            json.get("party_details").get("organisation_details").get("organisation_name").asText());
```
Aneesah33, line 71: Looks like nested OrganisationDetails may still need the legacy JSON names here, e.g. organisation_name. Could we check this and the other nested party detail fields/aliases too?

src/test/java/uk/gov/hmcts/opal/mapper/request/UpdateMinorCreditorAccountRequestMapperTest.java
```diff
@@ -0,0 +1,159 @@
...
+        assertNotNull(mapped.getPartyDetails());
+        assertEquals("99008", mapped.getPartyDetails().getPartyId());
+        assertEquals(false, mapped.getPartyDetails().getOrganisationFlag());
+        assertNotNull(mapped.getPartyDetails().getOrganisationDetails());
+        assertEquals("Updated Ltd", mapped.getPartyDetails().getOrganisationDetails().getOrganisationName());
+        assertNotNull(mapped.getPartyDetails().getOrganisationDetails().getOrganisationAliases());
+        assertEquals(1, mapped.getPartyDetails().getOrganisationDetails().getOrganisationAliases().length);
+        assertEquals(
+            "ORG-1",
+            mapped.getPartyDetails().getOrganisationDetails().getOrganisationAliases()[0].getAliasId()
```
Aneesah33, line 87: The legacy schema expects aliases to serialize with legacy JSON names too, e.g. organisation_aliases[].alias_id, sequence_number, organisation_name, and individual_aliases[].alias_id / sequence_number. This test confirms the Java values are mapped, but could the serialization test also assert the JSON field names for aliases?

src/test/java/uk/gov/hmcts/opal/mapper/request/UpdateMinorCreditorAccountRequestMapperTest.java
```diff
@@ -0,0 +1,159 @@
...
+        );
+        assertEquals((short) 7,
+            mapped.getPartyDetails().getOrganisationDetails().getOrganisationAliases()[0].getSequenceNumber());
+        assertNotNull(mapped.getPartyDetails().getIndividualDetails());
+        assertEquals("Ms", mapped.getPartyDetails().getIndividualDetails().getTitle());
+        assertEquals("Creditor", mapped.getPartyDetails().getIndividualDetails().getFirstNames());
```
Aneesah33, line 93: For firstNames snake_case would serialize this as first_names, but the linked legacy schema expects forenames. Could we make sure the serialization test covers that case too?

src/main/java/uk/gov/hmcts/opal/dto/legacy/LegacyUpdateMinorCreditorAccountRequest.java
HennaParmar: Thanks - I’ve updated the legacy DTO serialization so the outbound request uses the expected legacy field names rather than camelCase. I added the JSON naming/field annotations and then covered it with a serialization test against LegacyUpdateMinorCreditorAccountRequest.

src/test/java/uk/gov/hmcts/opal/dto/legacy/LegacyUpdateMinorCreditorAccountRequestTest.java
HennaParmar: Thanks - I’ve now covered the nested organisation details serialization as well, including organisation_name, so the test is asserting the legacy JSON shape rather than just the Java object values.

src/test/java/uk/gov/hmcts/opal/mapper/request/UpdateMinorCreditorAccountRequestMapperTest.java
HennaParmar: Good catch. I’ve extended the serialization coverage to include aliases as well, so we now assert the legacy field names for organisation_aliases and individual_aliases, including fields like alias_id and sequence_number.

src/test/java/uk/gov/hmcts/opal/mapper/request/UpdateMinorCreditorAccountRequestMapperTest.java
HennaParmar: Thanks - the serialization coverage now includes that case too, so we verify it serializes to the legacy schema field name forenames rather than a default snake_case conversion.

src/main/java/uk/gov/hmcts/opal/dto/legacy/LegacyUpdateMinorCreditorAccountRequest.java
HennaParmar: Updated same as above

Aneesah33 (approved):

PR #1787, opened 2026-05-18
Po 1986 get minor creditor opal

HennaParmar:

https://tools.hmcts.net/jira/browse/PO-1986

### Change description ###
Introduce new API to retrieve Minor Creditor Account details  GET: /minor-creditor-accounts/{id}

```
[ ] Yes
[ x ] No
```


thanhdat-nguyen-hmcts:
I noticed my ticket [PO-3762](https://tools.hmcts.net/jira/browse/PO-3762) requires putting a feature flag on the Get Minor Creditor Account API which is implemented in this PR. I will set my ticket as dependent on this PR being merged first.

src/main/java/uk/gov/hmcts/opal/controllers/MinorCreditorApiController.java
```diff
@@ -22,12 +24,19 @@
 public class MinorCreditorApiController implements MinorCreditorApi {
 
     private final MinorCreditorService minorCreditorService;
+    private final NativeWebRequest request;
```
Aneesah33: Do we still need to store NativeWebRequest on the controller here?

This was removed in PO-1991 after the thread-safety concern, and the GET path can now use UserStateService/Spring Security context instead of reading request headers directly.

src/main/java/uk/gov/hmcts/opal/controllers/MinorCreditorApiController.java
```diff
@@ -22,12 +24,19 @@
...
+        return Optional.ofNullable(request);
+    }
 
     @Override
     public ResponseEntity<MinorCreditorAccountResponseMinorCreditor> getMinorCreditorAccount(Long id) {
         log.debug(":GET:getMinorCreditorAccount: id={}", id);
 
-        MinorCreditorAccountResponse result = minorCreditorService.getMinorCreditorAccount(id);
+        MinorCreditorAccountResponse result =
+            minorCreditorService.getMinorCreditorAccount(id, request.getHeader("Authorization"));
```
Aneesah33: Should this call the no-header service method instead?

PO-1991 moved this GET flow towards resolving the authenticated user through Spring Security context, so this could be getMinorCreditorAccount(id) rather than passing the raw Authorization header through.

src/main/java/uk/gov/hmcts/opal/service/MinorCreditorService.java
```diff
@@ -137,25 +149,50 @@ public MinorCreditorAccountResponse updateMinorCreditorAccount(
...
+            businessUnitId,
+            hasBusinessUnitUser,
+            canViewBacs,
+            summariseBusinessUnits(userState));
+
+        if (canViewBacs) {
+            return response;
         }
+
+        response.getPayment().setPayByBacs(null);
```
Aneesah33: Is it safe to null this? The OpenAPI common schema marks pay_by_bacs as required, so returning null here may make the no-BACS-permission response non-schema-compliant.

Should we only redact the bank details fields and keep pay_by_bacs / hold_payment?

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalMinorCreditorIntegrationTest.java
```diff
@@ -277,6 +277,46 @@ void getHeaderSummary_authenticatedWithoutPermission_returns403() throws Excepti
...
 
+    @Test
+    void getMinorCreditorAccount_success_withBacsPermission_returnsBacsFields() throws Exception {
+        super.getMinorCreditorAccount_success_withBacsPermission_returnsBacsFields(log);
+    }
+
+    @Test
+    void getMinorCreditorAccount_success_withoutBacsPermission_redactsBacsFields() throws Exception {
+        super.getMinorCreditorAccount_success_withoutBacsPermission_redactsBacsFields(log);
+    }
```
Aneesah33, line 371: There is a shared helper called getMinorCreditorAccount_withBacsPermissionInDifferentBusinessUnit_redactsBacsFields in MinorCreditorControllerIntegrationTest, but I don’t think OpalMinorCreditorIntegrationTest currently calls it.

Could we add a @Test wrapper for it here so “View Creditor BACS” is also tested when the permission exists in a different business unit?

src/main/java/uk/gov/hmcts/opal/service/MinorCreditorService.java
```diff
@@ -17,19 +15,20 @@
...
 @RequiredArgsConstructor
 public class MinorCreditorService {
 
-    private static final String VIEW_CREDITOR_BACS_PERMISSION = "View Creditor BACS";
-
     private final MinorCreditorSearchProxy minorCreditorSearchProxy;
 
     private final UserStateService userStateService;
 
+    private final CreditorAccountRepository creditorAccountRepository;
```
Aneesah33: Is this still needed? I can’t see creditorAccountRepository being used in this service now that BACS filtering uses response.getBusinessUnitId().

Could we remove this dependency if it’s unused?

src/integrationTest/java/uk/gov/hmcts/opal/controllers/util/UserStateUtil.java
```diff
@@ -136,6 +136,22 @@ public static OpalJwtAuthenticationToken noFinesPermissionsToken() {
         return new OpalJwtAuthenticationToken(userState, FINES, getJwt(), emptySet(), null);
     }
 
+    public static OpalJwtAuthenticationToken permissionsToken(Short buid, FinesPermission... permissions) {
+        Map<Domain, DomainBusinessUnitUsers> domainsMap = new HashMap<>();
+        BusinessUnitUser businessUnitUser = BusinessUnitUser.builder()
```
Aneesah33: This helper makes sense for the SecurityContext tests. Minor tidy could it reuse the existing permissions(buid, permissionsFor(permissions)) helper when building the BusinessUnitUser, so the test user-state setup stays consistent?

src/main/java/uk/gov/hmcts/opal/controllers/MinorCreditorApiController.java
HennaParmar: Thanks - Removed. The controller no longer stores NativeWebRequest, and the GET path now uses the service method that resolves the authenticated user from Spring Security context.

src/main/java/uk/gov/hmcts/opal/controllers/MinorCreditorApiController.java
HennaParmar: Thanks - Updated. getMinorCreditorAccount(id) now calls the no-header service method, so the GET flow is aligned with the PO-1991 approach and no longer passes the raw Authorization header through.

src/main/java/uk/gov/hmcts/opal/service/MinorCreditorService.java
HennaParmar: Thanks - Good catch. I changed the redaction so we no longer null pay_by_bacs. We now keep pay_by_bacs and hold_payment intact and only redact the sensitive bank detail fields, which keeps the response aligned with the OpenAPI schema.

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalMinorCreditorIntegrationTest.java
HennaParmar: Thanks - added. OpalMinorCreditorIntegrationTest now includes a wrapper for the shared helper covering the case where View Creditor BACS exists in a different business unit, and the BACS fields are still redacted.

src/main/java/uk/gov/hmcts/opal/service/MinorCreditorService.java
HennaParmar: Thanks - Removed. CreditorAccountRepository was no longer used in MinorCreditorService, so I deleted the dependency and cleaned up the related unit test mock as well.

src/integrationTest/java/uk/gov/hmcts/opal/controllers/util/UserStateUtil.java
HennaParmar: Thanks I’ve updated permissionsToken(...) to reuse the existing permissions(buid, permissionsFor(...)) helper so the test user-state setup is now consistent with the other utility paths.

Aneesah33 (approved):

Alfred-Gillingham2 (approved):

PR #1789, opened 2026-05-18
Fixed issue that was causing master build to fail when running OpenAPIPublisherTest

Ben-Edwards-cgi:
Fixed issue that was causing master build to fail when running OpenAPIPublisherTest


CadeFaulkner (approved):

Sean-Reed-CGI (approved):
LGTM

PR #1810, opened 2026-05-26
PO-3368 WireMock-User-Service-API-Integration-Tests

shilhmcts:

https://tools.hmcts.net/jira/browse/PO-3368

### Change description ###

Refactored integration tests to stop mocking internal UserStateService/UserStateClientService behaviour.

Added shared WireMock setup and a reusable UserServiceStub so integration tests use the real security and user-state flow, while stubbing the external User Service HTTP responses.

Migrated integration tests to use valid JWTs and reusable User Service permission scenarios for authorised and forbidden cases.

```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/opal/AbstractIntegrationWithSecurityTest.java
```diff
@@ -41,36 +48,39 @@ static void beforeAll(@Autowired SecurityItProperties securityItProperties) thro
...
-                .withHeader("Content-Type", MediaType.APPLICATION_JSON_VALUE)
-                .withBody(V2_USER_STATE)));
-
-        RSASSASigner signer = new RSASSASigner(rsaKey);
-
-        validToken = validToken(rsaKey, signer, securityItProperties.getIssuerUri());
-        expiredToken = expiredToken(rsaKey, signer, securityItProperties.getIssuerUri());
+    @BeforeEach
+    void clearCachedUserState() {
+        redisTemplate.delete(USER_STATE_CACHE_PREFIX + "GfsHbIMt49WjQ");
```
Ben-Edwards-cgi: Can the subject be made as an global variable so it is not repeated

src/integrationTest/java/uk/gov/hmcts/opal/AbstractIntegrationWithSecurityTest.java
```diff
@@ -41,36 +48,39 @@ static void beforeAll(@Autowired SecurityItProperties securityItProperties) thro
...
         stubFor(get(jwksPath)
             .willReturn(aResponse()
                 .withHeader("Content-Type", MediaType.APPLICATION_JSON_VALUE)
                 .withBody(jwkResponse)));
+    }
 
-        stubFor(get("/opal/v2/users/0/state")
-            .willReturn(aResponse()
-                .withHeader("Content-Type", MediaType.APPLICATION_JSON_VALUE)
-                .withBody(V2_USER_STATE)));
```
Ben-Edwards-cgi, line 61: Why was this removed?

src/integrationTest/java/uk/gov/hmcts/opal/AbstractIntegrationTest.java
```diff
@@ -24,6 +26,13 @@
 @Import(IntegrationSecurityConfiguration.class)
 @ExtendWith(ZephyrAutomationExtension.class)
 @SuppressWarnings({"java:S6813", "SpringJavaInjectionPointsAutowiringInspection"})
+@EnableWireMock({
+    @ConfigureWireMock(
+        name = "user-service",
+        port = 0,
```
Ben-Edwards-cgi, line 34: Why is port 0?

in application.yaml we specific a port for each integration I believe

src/integrationTest/java/uk/gov/hmcts/opal/support/UserServiceStub.java
```diff
@@ -0,0 +1,338 @@
...
+
+    public static final String USER_STATE_PATH = "/v2/users/0/state";
+    public static final String V2_USER_STATE = """
+        {
+          "user_id" : 500000000,
+          "username" : "opal-test@HMCTS.NET",
+          "name" : "Pablo",
+          "status" : "ACTIVE",
+          "version" : 0,
+          "cache_name" : "USER_STATE_k9LpT2xVqR8m",
```
Ben-Edwards-cgi: k9LpT2xVqR8m should be the same as the subject on the JWT

src/integrationTest/java/uk/gov/hmcts/opal/support/UserServiceStub.java
```diff
@@ -0,0 +1,338 @@
...
+import static com.github.tomakehurst.wiremock.client.WireMock.get;
+import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
+
+import org.springframework.http.MediaType;
+import uk.gov.hmcts.opal.authorisation.model.FinesPermission;
+
+public final class UserServiceStub {
+
+    public static final String USER_STATE_PATH = "/v2/users/0/state";
+    public static final String V2_USER_STATE = """
```
Ben-Edwards-cgi: Could this setup Permissions dynamically based on the Permission Enum?
This is just so by default you have all perms (even if we add new ones).

Could use the ALL_PERMISSIONS from below?

src/integrationTest/java/uk/gov/hmcts/opal/support/UserServiceStub.java
```diff
@@ -0,0 +1,338 @@
...
+    private static final String USER_STATE_BY_ID_PATH_TEMPLATE = "/users/%d/state";
+    private static final String CREATE_MANAGE_DRAFT_ACCOUNTS_PERMISSION = """
+        [
+          {
+            "permission_id": 1,
+            "permission_name": "Create and Manage Draft Accounts"
+          }
+        ]
+        """;
+    private static final String ALL_PERMISSIONS = """
```
Ben-Edwards-cgi: Could these be dynamic instead of hard coded?

src/integrationTest/java/uk/gov/hmcts/opal/support/UserServiceStub.java
```diff
@@ -0,0 +1,338 @@
...
+            "permission_name": "Consolidate"
+          },
+          {
+            "permission_id": 14,
+            "permission_name": "Add and Remove payment hold"
+          }
+        ]
+        """;
+    private static final String NO_PERMISSIONS = "[]";
+    private static final String USER_STATE_BY_ID_TEMPLATE = """
```
Ben-Edwards-cgi: Why do we have USER_STATE_BY_ID_TEMPLATE and USER_STATE_TEMPLATE? 

Could the same one (with defaults) be used for both?

src/integrationTest/java/uk/gov/hmcts/opal/support/UserServiceStub.java
```diff
@@ -0,0 +1,338 @@
+package uk.gov.hmcts.opal.support;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.get;
+import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
+
+import org.springframework.http.MediaType;
+import uk.gov.hmcts.opal.authorisation.model.FinesPermission;
+
+public final class UserServiceStub {
```
Ben-Edwards-cgi, line 26: Instead of hard coding JSON in this class could we instead use standard  Objects (So build the UserStateV2 in code using builder patterns)
Then use Object Mapper to convert in into JSON?

This keeps it type safe and gives us better alerting to issues if the DTO changes.

This also means you can have one default builder that populates the minimum data, then you just extend upon this builder for each of your specific instances

src/integrationTest/java/uk/gov/hmcts/opal/support/UserServiceStub.java
```diff
@@ -0,0 +1,338 @@
...
+        """;
+
+    private UserServiceStub() {
+    }
+
+    public static void stubAuthorisedUser() {
+        stubUserWithAllPermissions(DEFAULT_BUSINESS_UNIT_ID);
+    }
+
+    public static void stubAuthorisedUser(String userStateJson) {
```
Ben-Edwards-cgi, line 49: Instead of passing in JSON could we have the UserStateV2 object?

Then we convert this into JSON?

Same for all places in this file.

JSON should only be used at the transaction boundary (I.e when we physically stub it), we should use Java native objects until then (This improves maintainability / usability)

src/integrationTest/java/uk/gov/hmcts/opal/AbstractIntegrationWithSecurityTest.java
shilhmcts: Removed the default `/opal/v2/users/0/state` stub from `AbstractIntegrationWithSecurityTest` because that base class should only set up shared security/JWT infrastructure. Each integration test now stubs the specific User Service response it needs via `UserServiceStub`, which makes permission scenarios explicit and avoids tests accidentally depending on shared WireMock state.

src/integrationTest/java/uk/gov/hmcts/opal/AbstractIntegrationTest.java
shilhmcts: By making the port 0 this would tell WireMock to start on any available port, avoiding clashes with the fixed `4553` test stub/legacy gateway port. `baseUrlProperties = "user.service.url"` then injects the actual dynamic WireMock URL into the app context, overriding the YAML default for integration tests.

.gitignore
```diff
@@ -33,6 +33,14 @@ target
...
 *.log
 
+### Shared OPAL Codex Skills ###
+.codex/skills/opal-fines-functional-test-review
+.codex/skills/opal-fines-functional-tests
+.codex/skills/opal-fines-integration-test-review
+.codex/skills/opal-fines-integration-tests
+.codex/skills/opal-java
+.codex/skills/review
+
```
RustyHMCTS: I'm not sure that our informal standard is to add codex skills into each individual repo, otherwise we risk duplication effort, and/or worse, having different standards in different repos.

I know this is also in the .gitignore, but you shouldn't have to exclude it in the first place ideally.

RustyHMCTS (changes requested):

src/integrationTest/java/uk/gov/hmcts/opal/AbstractIntegrationWithSecurityTest.java
shilhmcts: The test user subject is now defined once as `TEST_USER_SUBJECT`, and the Redis key is derived from it via `TEST_USER_STATE_CACHE_KEY`. `clearCachedUserState()` now uses the shared cache key

src/integrationTest/java/uk/gov/hmcts/opal/support/UserServiceStub.java
shilhmcts: The old mismatched `cache_name` value has been removed with the hard-coded JSON. The stub response is now generated from `UserStateV2`, and `k9LpT2xVqR8m` is no longer present. The JWT/cache subject used by the tests is centralised as `TEST_USER_SUBJECT`.

src/integrationTest/java/uk/gov/hmcts/opal/support/UserServiceStub.java
shilhmcts: hard-coded permission JSON has been removed; the default user state is now built from `UserStateV2` objects and derives all default permissions from `FinesPermission.values()` via `ALL_PERMISSIONS`, so new enum permissions are picked up automatically.

src/integrationTest/java/uk/gov/hmcts/opal/support/UserServiceStub.java
shilhmcts: This is now derived from `FinesPermission.values()` via `ALL_PERMISSIONS`

src/integrationTest/java/uk/gov/hmcts/opal/support/UserServiceStub.java
shilhmcts: The default `/v2/users/0/state` response is now generated from `UserStateV2` with shared defaults, and the `/users/{id}/state` helper now reuses the same shared default values for user name, business unit user id and permissions rather than maintaining a separate `USER_STATE_BY_ID_TEMPLATE`.

src/integrationTest/java/uk/gov/hmcts/opal/support/UserServiceStub.java
shilhmcts: The default user-state response is now built from `UserStateV2`/`BusinessUnitUser`/`Permission` objects and converted to JSON with the ObjectMapper via `TestUtil.toJsonString(...)`, rather than maintaining the response as hard-coded JSON

src/integrationTest/java/uk/gov/hmcts/opal/support/UserServiceStub.java
```diff
@@ -0,0 +1,348 @@
...
+                          permissionsJson);
+    }
+
+    private static String permissionObjectJson(FinesPermission permission) {
+        return """
+                    {
+                      "permission_id": %d,
+                      "permission_name": "%s"
+                    }""".formatted(permission.getId(), permission.getDescription());
+    }
```
Ben-Edwards-cgi: Should not be needed should use object approach

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalDefendantsPaymentTermsIntegrationTest.java
```diff
@@ -357,13 +325,11 @@ void test_Opal_AddPaymentTerms_IfMatchMissing() throws Exception {
 
     @Test
     @DisplayName("OPAL: Add Payment Terms - Schema validation error [@PO-1718]")
-    @JiraStory("PO-1718")
-    @JiraEpic("PO-977")
     void test_Opal_AddPaymentTerms_SchemaValidation() throws Exception {
         authoriseAllPermissions();
 
         HttpHeaders headers = new HttpHeaders();
```
Ben-Edwards-cgi, line 375: Why were these tests removed?

src/integrationTest/java/uk/gov/hmcts/opal/support/UserServiceStub.java
shilhmcts: Amended- UserServiceStub now builds the user-state responses using the relevant DTO/model objects and serializes them with the shared ObjectMapper helper

Ben-Edwards-cgi:
Closing as is replaced by: https://github.com/hmcts/opal-fines-service/pull/1900

PR #1842, opened 2026-06-03
PO-3435: Implement Cash List Report Service

shaunw-cgi:
https://tools.hmcts.net/jira/browse/PO-3435
BE - Cash List Report Service - Generate Report Object (New Internal Service)

### Change description ###
- Add Cash List report generation support to the reporting framework.
- Implement `CashListReportService` and `CashListReportData` for Cash List report data creation.
- Load and map Cash List report data based on F167 spec.
- Add temporary `convertReportDataToFileType(...)` stub implementation.
- Add unit and integration test coverage for Cash List report generation.

```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/opal/service/report/CashListReportServiceIntegrationTest.java
```diff
@@ -0,0 +1,161 @@
...
+import uk.hmcts.zephyr.automation.junit5.extension.ZephyrAutomationExtension;
+
+// Using a non-web context here to avoid starting MVC and reduce CI memory pressure
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.NONE)
+@ActiveProfiles("integration")
+@ContextConfiguration(classes = {TestContainerConfig.class})
+@Import(IntegrationSecurityConfiguration.class)
+@ExtendWith(ZephyrAutomationExtension.class)
+@Transactional
+class CashListReportServiceIntegrationTest {
```
marcmollins: why not extend AbstractIntegrationTest?

src/main/java/uk/gov/hmcts/opal/service/report/CashListPaymentLinkService.java
```diff
@@ -0,0 +1,50 @@
...
+@Component
+@RequiredArgsConstructor
+public class CashListPaymentLinkService {
+
+    private final DefendantAccountRepository defendantAccountRepository;
+    private final SuspenseItemRepository suspenseItemRepository;
+
+    public DefendantAccountEntity getDefendantAccount(PaymentInEntity payment) {
+        Long defendantAccountId = parseAssociatedRecordId(payment, DEFENDANT_ACCOUNTS.getLabel());
+        return defendantAccountRepository.findByDefendantAccountId(defendantAccountId)
```
marcmollins: there is a method in defendant account service for getting a defendantAccount, can we use that instead of calling the repository directly?

marcmollins (changes requested):

src/main/java/uk/gov/hmcts/opal/service/report/CashListPaymentLinkService.java
shaunw-cgi: Done, this now uses the DefendantAccountService to get the defendantAccounts rather than direct repository calls

src/integrationTest/java/uk/gov/hmcts/opal/service/report/CashListReportServiceIntegrationTest.java
shaunw-cgi: I originally had this extending AbstractIntegrationTest, but the pipeline was failing with an out of memory error during Spring context startup. I changed this test to use a non-web context as a workaround, which avoids the failure, but I agree it would be cleaner to keep using AbstractIntegrationTest if the OOM error can be solved another way

I can switch it back to extend AbstractIntegrationTest but I wasn’t sure what the best fix for the CI memory issue was

marcmollins: there is a temporary fix, so would prefer to use that instead of using your own workaround. the fix will hopefully be in master by the end of day

shaunw-cgi: I've modified this now to extend AbstractIntegrationTest again, so it can rely on a fix from master and not a local workaround

marcmollins (approved):
LGTM

stephenmagill (approved):
LGTM

PR #1854, opened 2026-06-04
P0-2251: Add Endpoint - Get Report Instances

ks21285-hmcts:

https://tools.hmcts.net/jira/browse/PO-2251

### Change description ###
Adds new endpoint GET `/report-instances` and includes
- controller and tests
- service and tests

```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/opal/controllers/ReportInstancesApiControllerIntegrationTest.java
```diff
@@ -0,0 +1,237 @@
...
+        @Test
+        @DisplayName("Get report instances - filter by report_id [@PO-2251]")
+        @JiraStory("PO-2251")
+        @JiraEpic("PO-2248")
+        void getReportInstances_whenFilteredByReportId_returnsMatchingInstances() throws Exception {
+            mockMvc.perform(get(URL_BASE)
+                    .param("report_id", REPORT_ID))
+                .andExpect(status().isOk())
+                .andExpect(jsonPath("$").isArray())
+                .andExpect(jsonPath("$.length()").value(3));
```
sarahtw-cgi: For these integration tests could you check on the returned object that it has been correctly filtered.
ie. here: `.andExpect(jsonPath("$[0].report_id").value("it_report_instances"))`

ks21285-hmcts: Good point, I have added them now

src/main/java/uk/gov/hmcts/opal/mapper/ReportInstanceMapper.java
```diff
@@ -0,0 +1,89 @@
...
+        List<SupportedFileType> types) {
+        if (types == null || types.isEmpty()) {
+            return List.of();
+        }
+        return types.stream()
+            .map(type -> ReportInstanceListReportsInner.SupportedFileTypesEnum.fromValue(type.name()))
+            .toList();
+    }
+
+    default List<BusinessUnitSummaryCommon> mapBusinessUnits(List<Long> businessUnitIds) {
```
Ben-Edwards-cgi: Can you check with @jacobgardner1999  please to see if there is a cleaner way to do this mapping as I don't think these defaults are required

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
```diff
@@ -90,11 +105,58 @@ private String getDataAsJson(ReportDataInterface data) throws JsonProcessingExce
         );
     }
 
+
+    @Override
+    public List<ReportInstanceListReportsInner> searchReportInstances(
```
Ben-Edwards-cgi, line 152: We need to ensure the user has the correct report permission for any report instance returned.

I.e if the user does not have the operation enforcement report permission these report instances should not be returned

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
```diff
@@ -90,11 +105,58 @@ private String getDataAsJson(ReportDataInterface data) throws JsonProcessingExce
...
+        final LocalDate fromDate,
+        final LocalDate toDate,
+        final List<Integer> businessUnits,
+        final Integer userId,
+        final String reportId) {
+        log.debug("Searching for report instances for report id {}", reportId);
+
+        UserState userState = userStateService.checkForAuthorisedUser();
+
+        final Specification<ReportInstanceEntity> spec = ReportInstanceSpecs.build(fromDate, toDate, userId, reportId);
```
Ben-Edwards-cgi: If a user tries to filter on a report they do not have access to (I.e they do not have permission for this report an error show be thrown

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
```diff
@@ -90,11 +105,58 @@ private String getDataAsJson(ReportDataInterface data) throws JsonProcessingExce
...
+
+        List<ReportInstanceEntity> reportInstances = reportInstanceRepository.findAll(spec);
+        if (reportInstances.isEmpty()) {
+            log.debug("No report instances found for reportId: {}", reportId);
+            return List.of();
+        }
+
+        if (businessUnits != null) {
```
Ben-Edwards-cgi: Should check not null and not empty

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
```diff
@@ -90,11 +105,58 @@ private String getDataAsJson(ReportDataInterface data) throws JsonProcessingExce
...
+
+        List<Long> businessUnitIds = reportInstances.stream()
+            .flatMap(instance ->
+                instance.getBusinessUnit() == null ? Stream.empty() : instance.getBusinessUnit().stream())
+            .filter(Objects::nonNull)
+            .toList();
+
+        checkAnyBusinessUsersHavePermissionForAnyBusinessUnits(userState, businessUnitIds);
+
+        ReportEntity report = reportRepository.findById(reportId)
```
Ben-Edwards-cgi: report_id is optional so may not always be provided.
You could have 10 report instances returned all  with different associated report id's

src/main/java/uk/gov/hmcts/opal/util/PermissionUtil.java
```diff
@@ -54,4 +60,31 @@ public static  <B extends BusinessUnitRef> List<B> filterBusinessUnitsByPermissi
                     .toList();
             }).orElse(refData);
     }
+
+    public static boolean checkAnyBusinessUsersHavePermissionForAnyBusinessUnits(
+        UserState userState,
```
Ben-Edwards-cgi: Instead of passing the UserState can we use the UserStateV2 (and use UserStateService to fetch it within this method instead i.e
 public static boolean checkAnyBusinessUsersHavePermissionForAnyBusinessUnits(
       List<Long> businessUnitIds) {

UserStateV2 userState = userStateService.getUserStateFromSecurityContext

src/main/java/uk/gov/hmcts/opal/mapper/ReportInstanceMapper.java
ks21285-hmcts: Checked with @jacobgardner1999 and improved ( a very little only) and updated the code.

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
ks21285-hmcts: Now it checks every report type if the user has permission.

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
ks21285-hmcts: Now it throws exception if reportId is provided which has no permission

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
ks21285-hmcts: Done

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
ks21285-hmcts: Now it is optional, that means, if no reportId is provided, the report ids are collected from the instances, and check if any of the reportId has permission. Only those reports are inlucded with permission, if none of the report Ids have permission, exception thrown.

src/main/java/uk/gov/hmcts/opal/util/PermissionUtil.java
ks21285-hmcts: UserStateService has this method now, so currently still UserState is used, waiting for the UserStateV2 to be available

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
Ben-Edwards-cgi: If reportId is not provided, we need to ensure only report instances that the user has access to are returned which I don't think is currently done.

To do this we will need to grap all the report_id's the user has access too (based on the associated perm) and filter on this

src/main/java/uk/gov/hmcts/opal/service/report/util/ReportInstanceUtil.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import org.springframework.security.access.AccessDeniedException;
+import uk.gov.hmcts.opal.entity.ReportEntity;
+import uk.gov.hmcts.opal.entity.ReportInstanceEntity;
+import uk.gov.hmcts.opal.repository.ReportRepository;
+import uk.gov.hmcts.opal.service.UserStateService;
+
+public class ReportInstanceUtil {
```
Ben-Edwards-cgi: This should be a service instead of a static class this way we can import the services it uses such as UserStateService

This voids the needs to pass this into the static each time

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
Ben-Edwards-cgi: We should only return items from the database that the user has access too else they may get 1000s of results returned that we then need to process down.

if no items are returned an empty array should be returned (as they made a valid search just there was not data found for there filters)

src/main/java/uk/gov/hmcts/opal/service/report/util/ReportInstanceUtil.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private ReportInstanceUtil() {
+        // utility
+    }
+
+    public static List<ReportInstanceEntity> filterByPermittedBusinessUnits(UserStateService userStateService,
```
Ben-Edwards-cgi: This should be filtered in the call to the database instead of in the Java layer

src/main/java/uk/gov/hmcts/opal/service/report/util/ReportInstanceUtil.java
```diff
@@ -0,0 +1,87 @@
...
+        return reportInstances.stream()
+            .filter(ri -> ri.getBusinessUnit() == null
+                || ri.getBusinessUnit().stream()
+                .filter(Objects::nonNull)
+                .anyMatch(effectiveBusinessUnitIds::contains))
+            .toList();
+
+    }
+
+    public static Map<ReportInstanceEntity, ReportEntity> filterByReportIds(ReportRepository reportRepository,
```
Ben-Edwards-cgi: This should be filtered in the database call instead of in the java layer

src/main/java/uk/gov/hmcts/opal/service/report/util/ReportInstanceUtil.java
```diff
@@ -0,0 +1,87 @@
...
+        return reportInstances.stream()
+            .filter(instance -> instance.getReportId() != null)
+            .filter(instance -> reportsById.containsKey(instance.getReportId()))
+            .collect(Collectors.toMap(
+                instance -> instance,
+                instance -> reportsById.get(instance.getReportId())
+            ));
+    }
+
+    public static Boolean hasReportPermission(ReportRepository reportRepository, UserStateService userStateService,
```
Ben-Edwards-cgi: This should be filtered in the database call instead of in the java layer

src/main/java/uk/gov/hmcts/opal/service/UserStateService.java
```diff
@@ -25,7 +30,7 @@ public class UserStateService {
     private final UserStateMapper userStateMapper;
 
     /**
```
Sean-Reed-CGI, line 33: Accidental edit here?

src/main/java/uk/gov/hmcts/opal/service/UserStateService.java
```diff
@@ -75,4 +80,58 @@ public UserStateV2 getUserStateFromSecurityContext() {
     public String getPreferredUsername(String authorization) {
         return extractPreferredUsername(authorization, tokenService);
     }
+
```
Sean-Reed-CGI, line 92: Unused private method

src/main/java/uk/gov/hmcts/opal/service/UserStateService.java
```diff
@@ -75,4 +80,58 @@ public UserStateV2 getUserStateFromSecurityContext() {
...
+        if (userState.getBusinessUnitUser() == null || userState.getBusinessUnitUser().isEmpty()) {
+            return "[]";
+        }
+
+        return userState.getBusinessUnitUser().stream()
+            .map(businessUnitUser -> String.valueOf(businessUnitUser.getBusinessUnitId()))
+            .sorted()
+            .collect(Collectors.joining(",", "[", "]"));
+    }
+
```
Sean-Reed-CGI: All the logic below probably belongs in the UserStateV2 class (in common lib)

src/main/java/uk/gov/hmcts/opal/service/report/util/ReportInstanceUtil.java
ks21285-hmcts: It is converted to service now : ReportInstanceSearchService

src/main/java/uk/gov/hmcts/opal/service/UserStateService.java
ks21285-hmcts: Oh, yes, I forgot to remove, fixed now

src/main/java/uk/gov/hmcts/opal/service/UserStateService.java
ks21285-hmcts: thanks, I removed now

src/main/java/uk/gov/hmcts/opal/service/UserStateService.java
ks21285-hmcts: Yes, but these also have Fines domain specific users. Shall  I add  these methods to common lib  and provide domain as parameter?

Sean-Reed-CGI: Yes, I think the domain will need to be passed as a parameter.

ks21285-hmcts: As we found these methods in authToken, I have removed them and used the authToken from SecurityUtil as you advised, thank you

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
```diff
@@ -140,13 +156,6 @@ private ReportInstanceEntity saveReportInstance(ReportInstanceEntity instance) {
         }
     }
```
Sean-Reed-CGI: private methods should be after public methods

ks21285-hmcts: I have moved the private methods to the public methods

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
```diff
@@ -156,11 +165,58 @@ private String getDataAsJson(ReportDataInterface data) throws JacksonException {
...
 
+    @Override
+    public List<ReportInstanceListReportsInner> searchReportInstances(
+        final LocalDate fromDate,
+        final LocalDate toDate,
+        final List<Integer> businessUnits,
+        final Integer userId,
+        final String reportId) {
+        log.debug("Searching for report instances");
+
```
Sean-Reed-CGI, line 159: If the user has selected a reportId there is no need to look up all permitted reports.

src/main/java/uk/gov/hmcts/opal/service/report/ReportInstanceSearchService.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    public ReportEntity findRequestedReportElseThrowError(List<ReportEntity> permittedReports, String reportId) {
+        if (reportId != null && !reportId.isBlank()) {
+            return permittedReports.stream()
+                .filter(report -> reportId.equals(report.getReportId()))
+                .findFirst()
+                .orElseThrow(() -> {
+                    if (reportRepository.existsById(reportId)) {
+                        throw new AccessDeniedException("User does not have permission for reportId: " + reportId);
```
Sean-Reed-CGI: Do we definitely want to be throwing these exceptions, should the user just get a an empty list of reports back if their search criteria brings nothing back?

ks21285-hmcts: Yes, after discussion with Ben last week, in these case we throw error

1. if reportId is provided, but it does not have permission
2. if businessUnitIds are provided, all of them must have permission, if not throw error

after going past these gates, 
If search criteria does not return anything, we return empty list

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
ks21285-hmcts: It now checks provided reportId first before getting all permitted reports.

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
```diff
@@ -132,12 +141,48 @@ public CreateReportInstanceResponseReports addReportInstance(CreateReportInstanc
...
+        Map<String, List<Long>> permittedReportForBusinessUnits =
+            reportInstanceSearchService.findPermittedReportForBusinessUnits(selectedReports, selectedBusinessUnitIds);
+
+        List<ReportInstanceEntity> reportInstances = new ArrayList<>();
+        permittedReportForBusinessUnits.forEach((reportIdKey, businessUnitIds) -> {
+            Specification<ReportInstanceEntity> spec =
+                ReportInstanceSpecs.build(fromDate, toDate, userId, reportIdKey, businessUnitIds);
+            reportInstances.addAll(reportInstanceRepository.findAll(spec));
+
```
Sean-Reed-CGI, line 177: Here we are querying the database once for every report.
So could be a lot of queries if a lot of reports. 
Is there no way we could get all the reports in a single hit?

src/main/java/uk/gov/hmcts/opal/service/report/ReportInstanceSearchService.java
```diff
@@ -0,0 +1,114 @@
...
+        return null;
+    }
+
+    public List<Long> findPermittedBusinessUnitIds() {
+        return getBusinessUnitUsers().stream()
+            .map(buUser -> buUser.getBusinessUnitId().longValue())
+            .distinct()
+            .toList();
+    }
+
```
Sean-Reed-CGI, line 53: 1. This method might be better named as validateSelectedBusinessUnitIds
2. It could look up the permitted business unit ids itself, rather than the GRS look these up and pass them in
3. It would be nice if the exception reported which business unit ids were the problem

src/main/java/uk/gov/hmcts/opal/mapper/ReportInstanceMapper.java
```diff
@@ -1,11 +1,75 @@
...
+    @Mapping(target = "generatedAt", source = "instance.createdTimestamp")
+    @Mapping(target = "requestedBy", source = "instance")
+    @Mapping(target = "name", expression = "java(mapReportName(instance, report))")
+    @Mapping(target = "businessUnits", expression = "java(mapBusinessUnits(instance.getBusinessUnit()))")
+    @Mapping(target = "status", source = "instance.generationStatus")
+    @Mapping(target = "numberOfRecords", source = "instance.noOfRecords")
+    @Mapping(target = "isDownloadable", expression = "java(calculateIsDownloadable(instance, report))")
+    @Mapping(target = "supportedFileTypes", source = "report.supportedFileTypes")
+    ReportInstanceListReportsInner toDto(ReportInstanceEntity instance, ReportEntity report);
+
```
Sean-Reed-CGI, line 35: Why do we need ReportEntity passed in? Can we just get it from the instance?

ks21285-hmcts: No, the ticket requires that the dto.name = reportInstance.reportName , if it's null, use the report.reportTitle
So both are needed

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
ks21285-hmcts: We discussed about this with Ben. Not all the permitted reports are allowed with all the permitted buisnessUnitIds. 
Lets say user has access to reports: r`eport_001`, `report_002`, and r`eport_003`
and has access to business_ids: `buid_001`, `buid_002`, and `buid_003`

we can not send db query with list of `List<reportIds>` and `List<buids>`, beause
report_001 is only allowed in `buid_001` and `buid_003`
report_002 is only allowed in `buid_001`
report_003 is only allowed in `buid_002`

So we have option to send 

1. max 20+ (reportIds)  requests i.e. map of reportId -> List of Buids
2. max 200+ (buids) requests i.e. map of buidl -> List of report ids

I choose 1 to reduce the number of calls.

src/main/java/uk/gov/hmcts/opal/service/report/ReportInstanceSearchService.java
ks21285-hmcts: I have made the changes now

src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java
Sean-Reed-CGI: OK, didn't realise it was discussed before. Thanks for explaining.

src/main/java/uk/gov/hmcts/opal/mapper/ReportInstanceMapper.java
Sean-Reed-CGI: Can't we just do reportIntance.getReport().getReportTitle in the mapping code?
Sorry if I am missing something!

ks21285-hmcts: Oh, actually, now we can. 
Until yesterday,  there was no ReportEntity in the ReportIncetanceEntity, it was reportInstance.reportId.
I have removed it now.

src/main/java/uk/gov/hmcts/opal/authorisation/model/FinesPermission.java
```diff
@@ -45,4 +46,11 @@ public static FinesPermission fromString(String value) {
             throw new IllegalArgumentException("Unknown FinesPermission: " + value, e);
         }
     }
+
```
Sean-Reed-CGI, line 49: Missing unit test

src/test/java/uk/gov/hmcts/opal/controllers/ReportInstancesApiControllerTest.java
```diff
@@ -0,0 +1,75 @@
...
+            },
+            () -> {
+                assert response.getBody() != null;
+                assertEquals(dto, response.getBody().getFirst());
+            },
+            () -> verify(genericReportService)
+                .searchReportInstances(FROM_DATE, TO_DATE, BUSINESS_UNITS, USER_ID, DEFAULT_REPORT_ID)
+        );
+    }
+
```
Sean-Reed-CGI: Don't think we need this second test. Only one code path so only one unit test required.

src/test/java/uk/gov/hmcts/opal/service/report/GenericReportServiceTest.java
```diff
@@ -42,95 +55,80 @@
...
         reportInstance.setReport(reportEntity);
+    }
 
+    void mockClock() {
         now = Instant.parse("2026-01-01T10:00:00Z");
-        Mockito.lenient().when(clock.instant()).thenReturn(now);
-        Mockito.lenient().when(clock.getZone()).thenReturn(ZoneOffset.UTC);
+        when(clock.instant()).thenReturn(now);
+        when(clock.getZone()).thenReturn(ZoneOffset.UTC);
     }
```
Sean-Reed-CGI, line 130: We shouldn't mock the Clock object, instead we can do:
` Clock.fixed(Instant.parse("2026-04-22T00:00:00Z"), ZoneOffset.UTC);`
Just noticed you are using existing pattern here, but would be good to fix

ks21285-hmcts: It was already there, I was putting it in a method. 
I have used the Clock.fixed now

src/test/java/uk/gov/hmcts/opal/service/report/ReportInstanceSearchServiceTest.java
```diff
@@ -0,0 +1,282 @@
...
+    @AfterEach
+    void clearSecurityContext() {
+        SecurityContextHolder.clearContext();
+    }
+
+    private void setAuthenticatedUserWithPermissions(
+        UserStateV2 userState,
+        uk.gov.hmcts.opal.authorisation.model.FinesPermission... permissions
+    ) {
+        OpalJwtAuthenticationToken authToken = org.mockito.Mockito.mock(OpalJwtAuthenticationToken.class);
```
Sean-Reed-CGI: Looks like this class wasn't imported properly. Also mocks can be declared with @Mock

src/test/java/uk/gov/hmcts/opal/service/report/ReportInstanceSearchServiceTest.java
```diff
@@ -0,0 +1,282 @@
...
+    private UserStateService userStateService;
+
+    @InjectMocks
+    private ReportInstanceSearchService reportInstanceSearchService;
+
+    @AfterEach
+    void clearSecurityContext() {
+        SecurityContextHolder.clearContext();
+    }
+
```
Sean-Reed-CGI, line 79: public methods before private methods

src/main/java/uk/gov/hmcts/opal/authorisation/model/FinesPermission.java
ks21285-hmcts: The unit test is added now

src/test/java/uk/gov/hmcts/opal/service/report/ReportInstanceSearchServiceTest.java
ks21285-hmcts: fixed now - private methods moved after all the public methods

src/test/java/uk/gov/hmcts/opal/service/report/ReportInstanceSearchServiceTest.java
ks21285-hmcts: Used @Mock now, and imported correctly now

Sean-Reed-CGI (approved):
LGTM

agoturi (approved):

stephenmagill (approved):

PR #1868, opened 2026-06-09
PO-2623-Amendment-Service

Elizabeth1420:
https://tools.hmcts.net/jira/browse/PO-2623

### Updates amendment auditing in order for the system to record who posted the change by ID and who they are by username. ###

Changes include:

- Added `postedByName` through the amendment service flow.
- Updated defendant account, party, payment terms, enforcement, and minor creditor call paths where audit finalise is used.
- Updated tests and integration expectations for the new `posted_by_name` behaviour.
- Fixed rebase/checkstyle issues after updating from master.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/opal/service/DefendantAccountPaymentTermsService.java
```diff
@@ -49,6 +49,7 @@ public AddPaymentCardRequestResponse addPaymentCardRequest(
                 defendantAccountId,
                 businessUnitId,
                 businessUnitUserId,
```
Aneesah33: Should we derive businessUnitUserId from UserState here, like the other audit flows, instead of passing the header value straight through?

src/main/resources/db/migration/ddl/V1_51__add_posted_by_name_to_audit_finalise.sql
Aneesah33, line 1: This is a db migration file and should not be in the backend changes. This has been implemented by the db work in PO-2596.

src/main/java/uk/gov/hmcts/opal/service/DefendantAccountPaymentTermsService.java
Elizabeth1420: Updated DefendantAccountPaymentTermsService to derive the business unit user ID from UserState rather than passing through the header value.

src/main/resources/db/migration/ddl/V1_51__add_posted_by_name_to_audit_finalise.sql
Elizabeth1420: I have removed the backend DB migration file

Aneesah33 (approved):
LGTM

Alfred-Gillingham2 (approved):

PR #1893, opened 2026-06-10
PO-2137 Add legacy major creditor at-a-glance endpoint

Aneesah33:

https://tools.hmcts.net/jira/browse/PO-2137

### Change description ###

- Adds legacy support for GET /major-creditor-accounts/{id}/at-a-glance
- Sends LIBRA.get_major_creditor_account_at_a_glance request payloads to legacy GOB
- Maps legacy XML responses to the OpenAPI response shape and returns ETag from creditor_account_version
- Enforces Search and View Accounts permission in any BU
- Adds controller, service, mapper, and integration coverage for the new legacy endpoint
- Adds matching WireMock mapping and XML stub in opal-legacy-db-stub 

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/opal/mapper/legacy/GetMajorCreditorAccountAtAGlanceResponseLegacyMapper.java
```diff
@@ -0,0 +1,27 @@
...
+    unmappedTargetPolicy = ReportingPolicy.IGNORE,
+    builder = @Builder(disableBuilder = true)
+)
+public interface GetMajorCreditorAccountAtAGlanceResponseLegacyMapper {
+
+    GetMajorCreditorAccountAtAGlanceResponse toOpal(GetMajorCreditorAccountAtAGlanceLegacyResponse legacy);
+
+    GetMajorCreditorAccountAtAGlanceResponse.MajorCreditor toOpal(MajorCreditorLegacy legacy);
+
+    @Mapping(target = "line1", source = "line1")
```
Sean-Reed-CGI: If target and source are the same no need to mention I think.

Aneesah33: Applied

src/integrationTest/java/uk/gov/hmcts/opal/controllers/LegacyMajorCreditorAccountAtAGlanceIntegrationTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+    private static final String AUTH_HEADER = "Bearer some_value";
+    private static final String URL = "/major-creditor-accounts/{id}/at-a-glance";
+
+    @MockitoBean
+    private UserStateService userStateService;
+
+    @MockitoSpyBean
+    private GatewayService gatewayService;
+
```
Sean-Reed-CGI: Ideally we shouldn't be using any mocks in the integration tests. We can't avoid mocking UserStateService  currently (sadly) but I suspect we don't need to use  @MockitoSpyBean for the GatewayService.

Aneesah33: That is right try to make it of minimum use, i have removed mocks so the test stay focused on endpoint behaviour.

Sean-Reed-CGI (approved):
LGTM

Ju-Morgan (approved):
Reviewed, ALGTM.

PR #1899, opened 2026-06-11
PO-7193 Preserve last enforcement and return result responses in enforcement flow

Aneesah33:

https://tools.hmcts.net/jira/browse/PO-7193

### Change description ###

- persist enforcement result_responses in the keyed format expected by enforcement status mapping
- preserve last_enforcement_action when payment_terms is included in POST /defendant-accounts/{id}/enforcements
- keep standalone addPaymentTerms behaviour unchanged and use an enforcement-specific path for the preserve case
- add unit and integration coverage for the merged enforcement + payment terms flow

```
[ ] Yes
[x] No
```


ShanCGI (approved):
Looks good to me

stephenmagill (approved):

PR #1913, opened 2026-06-15
PO-7222 Update operational report permissions 

HennaParmar:
https://tools.hmcts.net/jira/browse/PO-7222

### Change description ###
Fixed GET /reports/{id} returning 403 for the two operational reports by populating the missing report permission metadata in seeded data.
A new DB migration sets the permission for operational_report_enforcement and operational_report_payment to SEARCH_AND_VIEW_ACCOUNTS, which allows authorised users to retrieve those report definitions through the normal reports API path.
Integration coverage was updated to verify that both operational reports now return 200 OK for authorised users and still return 403 Forbidden when the user does not have the required permission.

```
[ ] Yes
[ x ] No
```


Sean-Reed-CGI (approved):
LGTM

Aneesah33 (approved):

vibhu-HMCTS (approved):

tonymccallion-hmcts (approved):
DB SQL script was previously added but has now been removed. Approving for DB team, assuming no SQL script will be added to the PR after approval.

PR #1919, opened 2026-06-16
Po 7280 missing 'allow additional action' flag

HennaParmar:
https://tools.hmcts.net/jira/browse/PO-7280

### Change description ###
Changes Made:

ResultEntity.java - Added 5 missing database columns:
allowPaymentTerms (Boolean)
allowAdditionalAction (Boolean) - Primary field from ticket
generatesWarrant (boolean)
requiresLja (Boolean)

ResultDto.java - Added 6 missing fields (including manualEnforcement which existed in entity but not DTO):
allowPaymentTerms
allowAdditionalAction 
generatesWarrant
requiresLja
manualEnforcement
enfNextPermittedActions

ResultControllerTest.java - Updated test to include new fields

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/opal/dto/ResultDto.java
+++ b/src/main/java/uk/gov/hmcts/opal/dto/ResultDto.java
@@ -85,4 +85,22 @@ public class ResultDto {
 
     @JsonProperty("requires_employment_data")
     private Boolean requiresEmploymentData;
+
+    @JsonProperty("allow_payment_terms")
+    private Boolean allowPaymentTerms;
+
+    @JsonProperty("allow_additional_action")
+    private Boolean allowAdditionalAction;
+
+    @JsonProperty("generates_warrant")
+    private boolean generatesWarrant;
+
+    @JsonProperty("requires_lja")
+    private Boolean requiresLja;
+
+    @JsonProperty("manual_enforcement")
+    private boolean manualEnforcement;
+
+    @JsonProperty("enf_next_permitted_actions")
+    private String enfNextPermittedActions;
 }
--- a/src/main/java/uk/gov/hmcts/opal/dto/ResultDto.java
+++ b/src/main/java/uk/gov/hmcts/opal/dto/ResultDto.java
@@ -90,6 +90,7 @@ public class ResultDto {
     private Boolean allowPaymentTerms;
 
     @JsonProperty("allow_additional_action")
+    @JsonInclude(JsonInclude.Include.ALWAYS)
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/opal/dto/ResultDto.java
```diff
@@ -85,4 +85,22 @@ public class ResultDto {
 
     @JsonProperty("requires_employment_data")
     private Boolean requiresEmploymentData;
+
+    @JsonProperty("allow_payment_terms")
+    private Boolean allowPaymentTerms;
+
+    @JsonProperty("allow_additional_action")
```
Aneesah33, line 92: allow_additional_action is added here, but ResultDto still uses @JsonInclude(JsonInclude.Include.NON_NULL). Since results.allow_additional_action is nullable in the db schema and this field is a Boolean, it can still be left out of the GET /results/{resultId} response when the DB value is null.

src/test/java/uk/gov/hmcts/opal/controllers/ResultControllerTest.java
```diff
@@ -68,6 +68,11 @@ void testGetResultDto_Success_WithSimpleJsonExpectation() {
             .resultTitleCy("Welsh Title")
             .resultType("TYPE1")
             .active(true)
+            .allowAdditionalAction(true)
```
Aneesah33, line 71: Ideally this should be covered by mapper/integration tests rather than a mocked controller test. At the moment this only proves the controller returns a mocked ResultDto, not that the real GET /results/{resultId} response includes these fields.

HennaParmar: Thanks I've now added proper mapper and integration tests that prove the real API response includes these fields, not just mocked data.

src/main/java/uk/gov/hmcts/opal/dto/ResultDto.java
HennaParmar: Thanks, good catch. I’ve updated ResultDto so allow_additional_action is explicitly included even when the mapped Boolean value is null, by adding field-level @JsonInclude(JsonInclude.Include.ALWAYS). I’ve also added a DTO serialization test to assert the response JSON contains "allow_additional_action": null when the value is null.

Aneesah33 (approved):
LGTM

Nipa1410 (approved):

PR #1920, opened 2026-06-16
PO-2256 Add detailed operation report by payment

sarahtw-cgi:
https://tools.hmcts.net/jira/browse/PO-2256

### Change description ###
Add report generation code for the detailed Operation by Payment report.

Restructured where the operation report code sits for better organisation. This was done in the first commit: https://github.com/hmcts/opal-fines-service/pull/1920/changes/e9619928cbb4d27f75fcc80751988bb48916fd39

Logic is in the subsequent commit/commits.

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/opal/service/report/operation/OperationReportByPaymentService.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+        return detailedResultMapper.map(accounts);
+    }
+
+    private List<DefendantAccountEntity> filterByLatestEnforcementPayment(
+        List<DefendantAccountEntity> accounts,
+        OperationReportByPaymentFiltersDto filters
+    ) {
+        boolean paymentMade = Boolean.TRUE.equals(filters.getIsPaymentMade());
+        return accounts.stream()
```
ShanCGI: Suggestion: Avoid per-account queries in the payment report filters

Problem: For `sinceLastEnforcementAction` and `isWithRegf`, the service loads all commonly matching accounts and then performs enforcement and transaction lookups inside a stream for each account.

Why: Operational reports can span multiple business units and large account populations. This introduces at least two extra database round trips per candidate account before row mapping starts, so the branch can degrade badly as soon as the common filters are broad. The TDIA classifies report creation as a backend operation and the shared review guidance calls out N+1 queries as high risk.

src/main/java/uk/gov/hmcts/opal/service/report/operation/OperationReportByPaymentService.java
```diff
@@ -0,0 +1,139 @@
...
+                account, filters.getSinceLastEnforcementAction()))
+            .toList();
+    }
+
+    private List<DefendantAccountEntity> filterByRegfPayment(
+        List<DefendantAccountEntity> accounts,
+        OperationReportByPaymentFiltersDto filters
+    ) {
+        boolean paymentMade = Boolean.TRUE.equals(filters.getIsPaymentMade());
+        return accounts.stream()
```
ShanCGI, line 78: See comment in filterByLatestEnforcementPayment(...) L64 above.

src/main/java/uk/gov/hmcts/opal/service/report/operation/OperationReportByPaymentService.java
```diff
@@ -0,0 +1,139 @@
...
+        OperationReportByPaymentFiltersDto filters = readFilters(reportInstance);
+        validator.validate(filters);
+        List<DefendantAccountEntity> accounts = defendantAccountRepository.findAll(
+            OperationReportSpecs.accountFiltersSpec(filters),
+            Sort.by(ACCOUNT_NUMBER)
+        );
+        if (filters.getSinceLastEnforcementAction() != null) {
+            return detailedResultMapper.map(filterByLatestEnforcementPayment(accounts, filters));
+        }
+        if (Boolean.TRUE.equals(filters.getIsWithRegf())) {
```
ShanCGI: Reject `isWithRegf=false` instead of treating it as the selected include-account mode

Problem: The validator increments the include-account filter count whenever `isWithRegf` is non-null, but the service only runs REGF logic when `Boolean.TRUE.equals(filters.getIsWithRegf())`.

Why: If report parameters contain `"isWithRegf": false` with no `sinceDate` or `sinceLastEnforcementAction`, validation passes even though no include-account option is actually selected. The service then falls through to `detailedResultMapper.map(accounts)` after the broad payment-made predicate, producing an untethered any-payment/no-payment-ever report instead of one of the three required modes.

src/main/java/uk/gov/hmcts/opal/service/report/operation/OperationReportByPaymentValidator.java
```diff
@@ -0,0 +1,48 @@
...
+    private void validateIncludeAccounts(OperationReportByPaymentFiltersDto filters,
+        List<String> validationErrors) {
+        int count = 0;
+        if (filters.getSinceLastEnforcementAction() != null) {
+            count++;
+        }
+        if (filters.getSinceDate() != null) {
+            count++;
+        }
+        if (filters.getIsWithRegf() != null) {
```
ShanCGI: See comment in `OperationReportByPaymentService.generateReportData(...)` L53

src/main/java/uk/gov/hmcts/opal/service/report/operation/OperationReportByPaymentService.java
```diff
@@ -0,0 +1,139 @@
...
+            .filter(account -> {
+                EnforcementEntity enforcement =
+                    enforcementRepository.findTopByDefendantAccountIdAndResultIdOrderByPostedDateAsc(
+                        account.getDefendantAccountId(),
+                        ResultId.REGF.name()
+                    );
+                if (enforcement == null || enforcement.getPostedDate() == null) {
+                    return false;
+                }
+                boolean hasPaymentAfterRegf =
```
ShanCGI: Cutoff checks count non-payment transactions as payments.

Problem: `hasPaymentAfterEnforcement(...)` and the REGF path use `existsByDefendantAccountIdAndPostedDateGreaterThanEqual(...)`, whose repository method only filters by account id and posted date.

Why: The report definition is specifically “Payment Made by the Defendant”. The branch already models that as `associated_record_type = DEFENDANT_ACCOUNTS`, transaction type `PAYMNT` or `CHEQUE`, and status `C` or `P` in `DefendantTransactionSpecs.paymentMadeOnOrAfterDatePredicate(...)`; the direct repository existence check ignores all of that. A `CONSOL`, write-off, reversed/cancelled transaction, or other non-payment after the cutoff can make an account appear as “payment made” and exclude it from “payment not made”.

src/main/java/uk/gov/hmcts/opal/service/report/operation/OperationReportByPaymentService.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private boolean hasPaymentAfterEnforcement(
+        DefendantAccountEntity account,
+        EnforcementEntity enforcement
+    ) {
+        if (enforcement == null || enforcement.getPostedDate() == null) {
+            return false;
+        }
+        return defendantTransactionRepository.existsByDefendantAccountIdAndPostedDateGreaterThanEqual(
```
ShanCGI: Cutoff checks count non-payment transactions as payments.

See comment in `filterByRegfPayment(...)` above.

src/main/java/uk/gov/hmcts/opal/service/report/operation/OperationReportByPaymentService.java
```diff
@@ -0,0 +1,139 @@
...
+    @Override
+    public ReportId getReportId() {
+        return OP_PAYMENT;
+    }
+
+    @Override
+    public ReportDataInterface generateReportData(ReportInstanceEntity reportInstance) {
+        OperationReportByPaymentFiltersDto filters = readFilters(reportInstance);
+        validator.validate(filters);
+        List<DefendantAccountEntity> accounts = defendantAccountRepository.findAll(
```
ShanCGI: Avoid prefilter since-last/REGF “payment not made” reports by any historic payment

Problem: `OperationReportSpecs.accountFiltersByPayment(...)` always adds `addIsPaymentMadeFilter(...)`, and when `sinceDate` is null that subquery checks for any payment at any time. For `sinceLastEnforcementAction` and `isWithRegf`, `OperationReportByPaymentService` then applies the actual cutoff in memory, but only after this broad prefilter has already removed accounts with earlier payments.

Why: Jira AC4 and AC6 require “payment not made” to include accounts with no defendant payment on or after the last enforcement/earliest REGF date. An account that paid before that cutoff but not after it should be returned, but the current DB predicate excludes it before the cutoff-specific check runs.

sarahtw-cgi: I've now made the paymentMadeFilter only apply to 'sinceDate' actions and REGF and sinceLastEnforcement are done with a Repository method

src/main/java/uk/gov/hmcts/opal/service/report/operation/OperationReportByPaymentService.java
sarahtw-cgi: This is now done with a call to a repository method

src/main/java/uk/gov/hmcts/opal/service/report/operation/OperationReportByPaymentService.java
sarahtw-cgi: I have created a "PaymentReportMode" parameter with 'WITH_REGF', 'SINCE_DATE', and 'SINCE_LAST_ENFORCEMENT' so false is not an option

src/main/java/uk/gov/hmcts/opal/service/report/operation/OperationReportByPaymentService.java
sarahtw-cgi: Repository method updated to now check for payment made

ShanCGI (approved):
Looks good.

Nipa1410 (approved):

PR #1939, opened 2026-06-18
Po 2254 get report instance endpoint

alex-reeves-cgi:

https://tools.hmcts.net/jira/browse/PO-2254

### Change description ###

Implement/add new endpoint for GET report instance entity by id

```
[ ] Yes
[X] No
```


src/test/java/uk/gov/hmcts/opal/service/report/GenericReportServiceTest.java
```diff
@@ -239,6 +254,68 @@ void generateReportInstanceContent_unableToSaveToDb_throwsException() {
...
+        when(reportInstanceRepository.findById(1L)).thenReturn(Optional.empty());
+
+        EntityNotFoundException exception = assertThrows(EntityNotFoundException.class,
+            () -> genericReportService.getReportInstance(1L));
+
+        assertEquals("Report instance not found with id: 1", exception.getMessage());
+    }
+
+    @Test
+    void getReportInstance_withoutBusinessUnit_returnsReportInstance() {
```
sarahtw-cgi: Should business units always be present?

src/main/java/uk/gov/hmcts/opal/mapper/common/JsonMapper.java
```diff
@@ -0,0 +1,30 @@
...
+        return objectMapper.readValue(jsonString, new TypeReference<>() {});
+    }
+
+    @Named("mapToJsonString")
+    default String mapToJsonString(Map<String, String> map) throws JsonProcessingException {
+        if (map == null) {
+            return null;
+        }
+        return objectMapper.writeValueAsString(map);
+    }
```
sarahtw-cgi: There's a JsonMapperHelper.class with similar logic. could these classes be merged?

sarahtw-cgi (commented):
Looks good. Left a couple of comments

src/test/java/uk/gov/hmcts/opal/service/report/GenericReportServiceTest.java
alex-reeves-cgi: I think that's true, but the db doesn't enforce not null on the business unit IDs field. This test is there for that plus coverage, I could remove the check in src/main/java/uk/gov/hmcts/opal/service/report/GenericReportService.java and the test, what do you think?

src/main/java/uk/gov/hmcts/opal/mapper/common/JsonMapper.java
alex-reeves-cgi: Good spot, yeah I'll use that. Looks like this outdates JsonMapperHelper but that's in master 1st

sarahtw-cgi (approved):

stephenmagill (approved):

PR #1949, opened 2026-06-19
PO-3858: BE enum aligntment with DB - PrintJobEntity

thanhdat-nguyen-hmcts:
https://tools.hmcts.net/jira/browse/PO-3858
https://tools.hmcts.net/jira/browse/PO-3843 (DB)

### Change description ###
Align BE code with DB enums for PrintJob entities

- Update the Entity classes to use enum value instead of string
- Update any logic within the application to use the enum instead of strings

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/opal/controllers/print/PrintRequestController.java
+++ b/src/main/java/uk/gov/hmcts/opal/controllers/print/PrintRequestController.java
@@ -17,7 +17,7 @@ import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
-import uk.gov.hmcts.opal.entity.print.PrintJob;
+import uk.gov.hmcts.opal.entity.print.PrintJobEntity;
 import uk.gov.hmcts.opal.service.print.AsyncPrintJobProcessor;
 import uk.gov.hmcts.opal.service.print.PrintService;
 
@@ -38,10 +38,10 @@ public class PrintRequestController {
 
     @PostMapping(value = "/enqueue-print-jobs", consumes = MediaType.APPLICATION_JSON_VALUE)
     @Operation(summary = "Enqueues print jobs for a batch of documents")
-    public ResponseEntity<String> enqueuePrintJobs(@RequestBody List<PrintJob> printJobs) {
-        log.debug(":POST:enqueuePrintJobs: received {} documents", printJobs.size());
+    public ResponseEntity<String> enqueuePrintJobs(@RequestBody List<PrintJobEntity> printJobEntities) {
+        log.debug(":POST:enqueuePrintJobs: received {} documents", printJobEntities.size());
 
-        UUID batchId = printService.savePrintJobs(printJobs);
+        UUID batchId = printService.savePrintJobs(printJobEntities);
 
         return ResponseEntity.ok().body(batchId.toString());
     }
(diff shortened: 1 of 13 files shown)


tonymccallion-hmcts (commented):
I have added the 1 db/migration script for PO-3843 on behalf of the DB team.

src/main/java/uk/gov/hmcts/opal/entity/print/PrintJobEntity.java
```diff
@@ -50,9 +50,10 @@ public class PrintJob {
     @Column(name = "doc_version", nullable = false, length = 20)
     private String docVersion;
 
+    @Builder.Default
     @Enumerated(EnumType.STRING)
     @Column(name = "status", nullable = false)
```
shaunw-cgi: This may require the annotations `@JdbcTypeCode(SqlTypes.NAMED_ENUM)` and 
`@Column(name = "status", nullable = false, columnDefinition = "t_print_job_status_enum")` otherwise JPA could still bind it as a varchar rather than the Postgres enum

shaunw-cgi (changes requested):

src/main/java/uk/gov/hmcts/opal/controllers/print/PrintRequestController.java
```diff
@@ -38,21 +38,21 @@ public class PrintRequestController {
 
     @PostMapping(value = "/enqueue-print-jobs", consumes = MediaType.APPLICATION_JSON_VALUE)
     @Operation(summary = "Enqueues print jobs for a batch of documents")
```
Sean-Reed-CGI, line 40: Probably we should be receiving a DTO here, rather than JPA entities

src/main/java/uk/gov/hmcts/opal/entity/print/PrintJobEntity.java
thanhdat-nguyen-hmcts: The annotations were added. It is surprising the tests did not fail in their absence.

src/main/java/uk/gov/hmcts/opal/controllers/print/PrintRequestController.java
thanhdat-nguyen-hmcts: @Ben-Edwards-cgi Could this be possibly spun off into a different ticket?

shaunw-cgi (approved):
LGTM

src/main/java/uk/gov/hmcts/opal/controllers/print/PrintRequestController.java
Ben-Edwards-cgi: We will need to update this in this ticket as we should avoid directly returning entities as we could expose information we never intended to which could lead to GDPR issues or general bugs in the code as we could send something the FE is not expecting.

thanhdat-nguyen-hmcts: I have replaced the entity with a DTO and created a mapper to convert from one to another

src/main/java/uk/gov/hmcts/opal/controllers/print/PrintRequestController.java
```diff
@@ -31,28 +33,32 @@ public class PrintRequestController {
...
-    public ResponseEntity<String> enqueuePrintJobs(@RequestBody List<PrintJob> printJobs) {
-        log.debug(":POST:enqueuePrintJobs: received {} documents", printJobs.size());
+    public ResponseEntity<String> enqueuePrintJobs(@RequestBody List<PrintJobDto> printJobDtos) {
+        log.debug(":POST:enqueuePrintJobs: received {} documents", printJobDtos.size());
 
-        UUID batchId = printService.savePrintJobs(printJobs);
+        List<PrintJobEntity> printJobEntities = printJobMapper.toEntities(printJobDtos);
+        UUID batchId = printService.savePrintJobs(printJobEntities);
```
Sean-Reed-CGI: Better to do the mapping in the service layer as we want minimal logic in the controller.

thanhdat-nguyen-hmcts: The changes were applied

Sean-Reed-CGI (approved):
LGTM

Alfred-Gillingham2 (approved):

cl-cgi (approved):
Reviewed and passed

PR #1981, opened 2026-06-25
PO-7287: Update length originator_name in defendant_account.

RustyHMCTS:
Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```


RustyHMCTS:
Fixed by PO-7287

PR #1960, opened 2026-06-22
PO-2434 Get enforcement account types endpoint

jordan-muscott-cgi:
### JIRA link ###
https://tools.hmcts.net/jira/browse/PO-2434

### Change description ###
GET endpoint for Enforcement Account Types reference data.
This db table was previously unused, and the db ticket to put data into that table has only just been merged.

**NOTE** - the was an existing enum called "EnforcementAccountType" which has been changed by the db team to be an extended enum that includes legacy values in addition to the OPAL values. I have had to change this existing enum to be called "EnforcementAccountTypeExtended" in order to reflect this change. Note this is not referenced by the db table we are querying, as our table contains the new slimmer version of the enum that just has OPAL values.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #1985, opened 2026-06-26
PO-2435 update enforcement account type

thomas-croasdale-cgi:
[### JIRA link (if applicable) ###](https://tools.hmcts.net/jira/browse/PO-2435)

Implemented api for updating enforcement account types.

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/opal/entity/enforcement/AccountType.java
```diff
@@ -0,0 +1,27 @@
+package uk.gov.hmcts.opal.entity.enforcement;
+
+import lombok.Getter;
+
+import java.util.stream.Stream;
+
+public enum AccountType {
```
jordan-muscott-cgi, line 6: This file will be completely replaced by the code from https://github.com/hmcts/opal-fines-service/pull/1960/changes

src/main/java/uk/gov/hmcts/opal/entity/enforcement/EnforcementAccountType.java
```diff
@@ -0,0 +1,28 @@
+package uk.gov.hmcts.opal.entity.enforcement;
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public enum EnforcementAccountType {
```
jordan-muscott-cgi, line 7: This file will be completely replaced by the file in https://github.com/hmcts/opal-fines-service/pull/1960/changes

src/main/java/uk/gov/hmcts/opal/entity/enforcement/EnforcementAccountTypeEntity.java
```diff
@@ -0,0 +1,66 @@
...
+import uk.gov.hmcts.opal.util.Versioned;
+
+@Entity
+@Table(name = "enforcement_account_types")
+@NoArgsConstructor
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+@Builder
+@Getter
+@Setter
+public class EnforcementAccountTypeEntity implements Versioned {
```
jordan-muscott-cgi, line 33: This file will be mostly replaced by the code in https://github.com/hmcts/opal-fines-service/pull/1960/changes

src/main/java/uk/gov/hmcts/opal/entity/LowHighValue.java
```diff
@@ -0,0 +1,28 @@
+package uk.gov.hmcts.opal.entity;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+import java.util.stream.Stream;
+
+public enum LowHighValue {
```
jordan-muscott-cgi, line 6: This file will be replaced by the code in https://github.com/hmcts/opal-fines-service/pull/1960/changes

src/main/java/uk/gov/hmcts/opal/mapper/EnforcementAccountTypeMapper.java
```diff
@@ -0,0 +1,23 @@
...
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import uk.gov.hmcts.opal.entity.enforcement.EnforcementAccountTypeEntity;
+import uk.gov.hmcts.opal.generated.model.EnforcementAccountTypeCommon;
+
+import java.util.Collection;
+import java.util.List;
+
+@Mapper(componentModel = "spring")
+public interface EnforcementAccountTypeMapper {
```
jordan-muscott-cgi, line 11: This file will be replaced by the code in https://github.com/hmcts/opal-fines-service/pull/1960/changes

src/main/java/uk/gov/hmcts/opal/repository/EnforcementAccountTypeRepository.java
```diff
@@ -0,0 +1,11 @@
+package uk.gov.hmcts.opal.repository;
+
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
+import org.springframework.stereotype.Repository;
+import uk.gov.hmcts.opal.entity.enforcement.EnforcementAccountTypeEntity;
+
+@Repository
+public interface EnforcementAccountTypeRepository extends JpaRepository<EnforcementAccountTypeEntity, Long>,
```
jordan-muscott-cgi, line 9: This file will be replaced by the code in https://github.com/hmcts/opal-fines-service/pull/1960/changes

src/main/java/uk/gov/hmcts/opal/service/EnforcementAccountTypesService.java
```diff
@@ -0,0 +1,80 @@
...
+                throw new ObjectOptimisticLockingFailureException("versionNumber", entity.getVersionNumber(),
+                                                                  "Version numbers do not match", null);
+            }
+
+            entity.setMinimumBalance(requestObject.getMinimumBalance());
+            entity.setVersionNumber(requestObject.getVersion().longValue() + 1);
+
+            entities.add(entity);
+        }
+        List<EnforcementAccountTypeEntity> updatedEntities = enforcementAccountTypeRepository.saveAll(entities);
```
jordan-muscott-cgi: If these are updates then a call to saveAll is not necessary.
If the service method is annotated with @Transactional then the JPA session will be flushed and the transaction committed.

src/main/java/uk/gov/hmcts/opal/service/EnforcementAccountTypesService.java
```diff
@@ -0,0 +1,80 @@
...
+@Service
+@Slf4j(topic = "opal.EnforcementAccountTypesService")
+@RequiredArgsConstructor
+public class EnforcementAccountTypesService {
+
+    private final EnforcementAccountTypeRepository enforcementAccountTypeRepository;
+    private final UserStateService userStateService;
+    private final EnforcementAccountTypeMapper enforcementAccountTypeMapper;
+
+    public PatchEnforcementAccountType200Response updateEnforcementAccountType(
```
jordan-muscott-cgi: missing the @Transactional annotation

src/main/java/uk/gov/hmcts/opal/service/EnforcementAccountTypesService.java
```diff
@@ -0,0 +1,80 @@
...
+            }
+
+            entity.setMinimumBalance(requestObject.getMinimumBalance());
+            entity.setVersionNumber(requestObject.getVersion().longValue() + 1);
+
+            entities.add(entity);
+        }
+        List<EnforcementAccountTypeEntity> updatedEntities = enforcementAccountTypeRepository.saveAll(entities);
+
+        var resp = new PatchEnforcementAccountType200Response();
```
jordan-muscott-cgi: Perhaps me being a bit fussy - but I'm wondering if a class with "200" in its name should be referenced in the service layer? Just because "200" is an http status which could be the concern of the API layer (ie the controller).

I did deliberate about this in my PR, and ended up returning the DTOs from the service, and then constructing the XXXX200 object in the controller.

src/main/java/uk/gov/hmcts/opal/service/opal/EnforcementAccountTypeService.java
```diff
@@ -0,0 +1,80 @@
...
+@Slf4j(topic = "opal.EnforcementAccountTypesService")
+@RequiredArgsConstructor
+public class EnforcementAccountTypesService {
+
+    private final EnforcementAccountTypeRepository enforcementAccountTypeRepository;
+    private final UserStateService userStateService;
+    private final EnforcementAccountTypeMapper enforcementAccountTypeMapper;
+
+    public PatchEnforcementAccountType200Response updateEnforcementAccountType(
+        List<PatchEnforcementAccountTypeRequestInner> request) {
```
jordan-muscott-cgi, line 44: Needs unit tests

src/main/java/uk/gov/hmcts/opal/controllers/EnforcementAccountTypesApiController.java
```diff
@@ -0,0 +1,43 @@
...
+import static uk.gov.hmcts.opal.util.FeatureFlags.RELEASE_1C_AUTO_ENFORCEMENT_CONFIG;
+import static uk.gov.hmcts.opal.util.FeatureFlags.RELEASE_1C_AUTO_ENFORCEMENT_CONFIG_ENABLED_PROPERTY;
+import static uk.gov.hmcts.opal.util.HttpUtil.buildResponse;
+
+@RestController
+@Slf4j(topic = "opal.EnforcementAccountTypesController")
+@RequiredArgsConstructor
+@Tag(name = "Enforcement Account Types Controller")
+public class EnforcementAccountTypesApiController implements EnforcementAccountTypesApi {
+
```
jordan-muscott-cgi, line 31: Prob needs a basic orchestration unit test

src/integrationTest/resources/db/insertData/insert_into_enforcement_account_types.sql
```diff
@@ -0,0 +1,13 @@
+-- Insert some test data
+DELETE FROM enforcement_account_types
+WHERE enforcement_account_type_id IN (1, 2, 3);
+
+INSERT INTO enforcement_account_types (enforcement_account_type_id, account_type, minimum_balance,
+                                       enforcement_account_type, account_type_path, version_number)
+VALUES
+  (1, 'COL', 100, 'COLH',
```
jordan-muscott-cgi: the IDs in this script may now conflict with the enforcement_account_types that are in master ?

thomas-croasdale-cgi: They do, because of the constraints on unique fields we aren't able to create any rows specifically for these tests. I have changed this script to get called after the tests, to ensure that these rows stay consistent with what's in the db.

src/main/java/uk/gov/hmcts/opal/service/EnforcementAccountTypesService.java
thomas-croasdale-cgi: I Agree, I've updated this method to return a list of dtos instead

src/main/java/uk/gov/hmcts/opal/service/EnforcementAccountTypesService.java
thomas-croasdale-cgi: Done

src/main/java/uk/gov/hmcts/opal/service/EnforcementAccountTypesService.java
thomas-croasdale-cgi: Done

src/main/java/uk/gov/hmcts/opal/service/opal/EnforcementAccountTypeService.java
thomas-croasdale-cgi: Done

src/main/java/uk/gov/hmcts/opal/controllers/EnforcementAccountTypesApiController.java
thomas-croasdale-cgi: Done

src/main/java/uk/gov/hmcts/opal/service/opal/EnforcementAccountTypesService.java
```diff
@@ -0,0 +1,78 @@
...
+                throw new UnprocessableException("Can not update enforcement account type minimum balance for a low "
+                    + "enforcement path");
+            }
+
+            if (requestObject.getMinimumBalance() != null
+                && Optional.of(requestObject.getMinimumBalance()).orElse(new BigDecimal(0)).longValue() < 0L) {
+                throw new UnprocessableException("Can not set minimum balance to a negative value");
+            }
+
+            if (entity.getVersionNumber() != requestObject.getVersion().longValue()) {
```
jordan-muscott-cgi: Sorry I've only just noticed that the existing codebase has a VersionUtil class to check the version and throw the Optimistic locking exception. I guess its probably best to stick with this convention.

src/main/java/uk/gov/hmcts/opal/service/opal/EnforcementAccountTypesService.java
```diff
@@ -0,0 +1,78 @@
...
+                throw new ObjectOptimisticLockingFailureException("versionNumber", entity.getVersionNumber(),
+                    "Version numbers do not match", null);
+            }
+
+            entity.setMinimumBalance(requestObject.getMinimumBalance());
+            entity.setVersionNumber(requestObject.getVersion().longValue() + 1);
```
jordan-muscott-cgi: Again sorry I've just noticed the existing codebase uses the entity manager to lock the entity and increase the version number.

tbh the code I've just looked at in the OpalDefendantAccountService does this by injecting the JPA entity manager into the service, which I really don't like doing - personally i think the EntityManager should remain encapsulated in the data access layer... but anyway probably worth looking at !

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypesServiceTest.java
```diff
@@ -0,0 +1,191 @@
...
+            when(authToken.hasPermission(FinesPermission.AUTO_ENFORCEMENT)).thenReturn(true);
+            securityUtil.when(SecurityUtil::getOpalJwtAuthenticationTokenForCurrentUser).thenReturn(authToken);
+            List<PatchEnforcementAccountTypeRequestInner> request = List.of(
+                PatchEnforcementAccountTypeRequestInner.builder()
+                    .id(1)
+                    .minimumBalance(null)
+                    .build()
+            );
+
+            when(repository.findById(any())).thenReturn(Optional.of(
```
jordan-muscott-cgi: rather than doing `findById(any())` would be a bit nice to use the id that you are expecting to be passed in

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypesServiceTest.java
```diff
@@ -0,0 +1,191 @@
...
+            when(authToken.hasPermission(FinesPermission.AUTO_ENFORCEMENT)).thenReturn(true);
+            securityUtil.when(SecurityUtil::getOpalJwtAuthenticationTokenForCurrentUser).thenReturn(authToken);
+            List<PatchEnforcementAccountTypeRequestInner> request = List.of(
+                PatchEnforcementAccountTypeRequestInner.builder()
+                    .id(1)
+                    .minimumBalance(new BigDecimal("-100"))
+                    .build()
+            );
+
+            when(repository.findById(any())).thenReturn(Optional.of(
```
jordan-muscott-cgi: same comment as above - but slightly nice if the findById was setup to work the the specific id

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypesServiceTest.java
```diff
@@ -0,0 +1,191 @@
...
+            securityUtil.when(SecurityUtil::getOpalJwtAuthenticationTokenForCurrentUser).thenReturn(authToken);
+            List<PatchEnforcementAccountTypeRequestInner> request = List.of(
+                PatchEnforcementAccountTypeRequestInner.builder()
+                    .id(1)
+                    .minimumBalance(new BigDecimal("100"))
+                    .version(5)
+                    .build()
+            );
+
+            when(repository.findById(any())).thenReturn(Optional.of(
```
jordan-muscott-cgi: same comment as above - be nicer not to use any() here

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypesServiceTest.java
```diff
@@ -0,0 +1,191 @@
...
+            securityUtil.when(SecurityUtil::getOpalJwtAuthenticationTokenForCurrentUser).thenReturn(authToken);
+            List<PatchEnforcementAccountTypeRequestInner> request = List.of(
+                mock(PatchEnforcementAccountTypeRequestInner.class)
+            );
+
+            when(repository.findById(any())).thenReturn(Optional.of(mock(EnforcementAccountTypeEntity.class)));
+
+            service.updateEnforcementAccountType(request);
+
+            verify(mapper).toDtos(any());
```
jordan-muscott-cgi: would be a bit better to verift the mapper has been called with a specfic list of objects rather than just any()

src/main/java/uk/gov/hmcts/opal/service/opal/EnforcementAccountTypesService.java
thomas-croasdale-cgi: Agreed, This has been changed to stick with the convention

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypesServiceTest.java
thomas-croasdale-cgi: Done

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypesServiceTest.java
thomas-croasdale-cgi: Done

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypesServiceTest.java
thomas-croasdale-cgi: Done

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypesServiceTest.java
thomas-croasdale-cgi: Done

src/integrationTest/resources/db/deleteData/delete_from_enforcement_account_types.sql
```diff
@@ -0,0 +1,13 @@
+-- Reset the test data
+DELETE FROM enforcement_account_types
```
jordan-muscott-cgi: This looks 100% fine for now... but thinking forward, what if we end up with other data that has foreign keys pointing to these rows?

Could we just run an update statement to set their values back instead?

src/main/java/uk/gov/hmcts/opal/service/opal/EnforcementAccountTypesService.java
thomas-croasdale-cgi: I've tried out your suggestions for this, and I still feel .setVersionNumber() is better. It is explicit in what it's doing, and allows us to create the DTOs without needing to fetch the new version number.

The EntityManager lock methods did increase the version number on the db, but the version number we have stored on the entity would not increase, meaning we'd then have to fetch it again before creating the DTOs which is added complexity.

Doing a repository.flush() had the side effect of not increasing the version number if nothing has changed on the entity which is not quite what we want here. It is also not as explicit in what it's actually doing.

src/main/java/uk/gov/hmcts/opal/entity/enforcement/AccountType.java
thomas-croasdale-cgi: This branch is now updated to include master, resolving this thread.

src/main/java/uk/gov/hmcts/opal/entity/enforcement/EnforcementAccountType.java
thomas-croasdale-cgi: This branch is now updated to include master, resolving this thread.

src/main/java/uk/gov/hmcts/opal/entity/enforcement/EnforcementAccountTypeEntity.java
thomas-croasdale-cgi: This branch is now updated to include master, resolving this thread.

src/main/java/uk/gov/hmcts/opal/entity/LowHighValue.java
thomas-croasdale-cgi: This branch is now updated to include master, resolving this thread.

src/main/java/uk/gov/hmcts/opal/mapper/EnforcementAccountTypeMapper.java
thomas-croasdale-cgi: This branch is now updated to include master, resolving this thread.

src/main/java/uk/gov/hmcts/opal/repository/EnforcementAccountTypeRepository.java
thomas-croasdale-cgi: This branch is now updated to include master, resolving this thread.

src/integrationTest/java/uk/gov/hmcts/opal/controllers/AbstractOpalEnforcementAccountTypesIntegrationTest.java
```diff
@@ -0,0 +1,52 @@
...
+    @MockitoBean
+    protected UserState userState;
+
+    @MockitoBean
+    protected AccessTokenService accessTokenService;
+
+    @BeforeEach
+    void setupUserState() {
+        Mockito.when(userState.anyBusinessUnitUserHasPermission(Mockito.any())).thenReturn(true);
+        Mockito.when(userStateService.checkForAuthorisedUser(Mockito.any())).thenReturn(userState);
```
jordan-muscott-cgi: I know its been done elsewhere in the code base, but in the integration tests we shouldn't be mocking anything which runs on the server at all, as it defeats the point of an integration test.

src/test/java/uk/gov/hmcts/opal/controllers/EnforcementAccountTypesApiControllerTest.java
```diff
@@ -52,4 +56,21 @@ void getEnforcementAccountTypes_ReturnsOkWithEmptyCollection() {
...
         );
     }
-}
+
+    @Test
+    void updateEnforcementAccountTypes_Success() {
+        List<EnforcementAccountTypeCommon> enfAccountTypes = List.of(
+            mock(EnforcementAccountTypeCommon.class)
+        );
+        when(service.updateEnforcementAccountType(any())).thenReturn(enfAccountTypes);
```
jordan-muscott-cgi: Would be better to set up mocks to work with exact parameters to make sure they are being called in exactly the correct  way rather than just using "`any()`"

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
```diff
@@ -73,4 +86,146 @@ public void getEnforcementAccountTypes_unauthorisedUser_shouldThrowPermissionsEx
...
+        try (MockedStatic<SecurityUtil> securityUtil = mockStatic(SecurityUtil.class)) {
+            when(authToken.hasPermission(FinesPermission.AUTO_ENFORCEMENT)).thenReturn(true);
+            securityUtil.when(SecurityUtil::getOpalJwtAuthenticationTokenForCurrentUser).thenReturn(authToken);
+            List<PatchEnforcementAccountTypeRequestInner> request = List.of(
+                mock(PatchEnforcementAccountTypeRequestInner.class)
+            );
+            EnforcementAccountTypeEntity mockEntity = mock(EnforcementAccountTypeEntity.class);
+            when(mockEntity.getVersion()).thenReturn(BigInteger.ZERO);
+
+            when(repository.findById(any())).thenReturn(Optional.of(mockEntity));
```
jordan-muscott-cgi: As per previous comment much better to set up mocks to expect exact params rather then "any()"

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
```diff
@@ -73,4 +86,146 @@ public void getEnforcementAccountTypes_unauthorisedUser_shouldThrowPermissionsEx
...
+            securityUtil.when(SecurityUtil::getOpalJwtAuthenticationTokenForCurrentUser).thenReturn(authToken);
+            List<PatchEnforcementAccountTypeRequestInner> request = List.of(
+                PatchEnforcementAccountTypeRequestInner.builder().id(1L).build()
+            );
+
+            when(repository.findById(eq(1L))).thenReturn(Optional.ofNullable(null));
+
+            assertThrows(EntityNotFoundException.class, () -> service.updateEnforcementAccountType(request));
+            verifyNoInteractions(mapper);
+            verify(repository).findById(any());
```
jordan-muscott-cgi: see previous comments about any() - although tbh you could prob get away with not even bothering with the verify here because if the call hadn't been made then the assertThrows wouldn't have passed anyway !

src/main/java/uk/gov/hmcts/opal/service/opal/EnforcementAccountTypesService.java
jordan-muscott-cgi: tbh 100% agree with you that this looks much tidier but think we should probably check what the agreed project convention for doing this is.

Also - just out of curiosity, do you know why we want the version number to increase even if nothing has changed?

thomas-croasdale-cgi: When I asked Ben Edwards about this on a different ticket, This is what he said.
"I don't think it matters either way for this
 
As we only use the version for optimistic locking

so if it does get increased there is no impact and if it does not there is also no impact as there has been no change"

src/integrationTest/java/uk/gov/hmcts/opal/controllers/AbstractOpalEnforcementAccountTypesIntegrationTest.java
thomas-croasdale-cgi: Removed as was unnecessary

src/test/java/uk/gov/hmcts/opal/controllers/EnforcementAccountTypesApiControllerTest.java
thomas-croasdale-cgi: Done

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
thomas-croasdale-cgi: Done

src/main/java/uk/gov/hmcts/opal/entity/enforcement/EnforcementAccountTypeEntity.java
```diff
@@ -32,21 +39,30 @@ public class EnforcementAccountTypeEntity {
...
     private BigDecimal minimumBalance;
 
     @Column
+    @Version
     private Long versionNumber;
+
+    @Override
+    public BigInteger getVersion() {
+        return Optional.ofNullable(versionNumber).map(BigInteger::valueOf).orElse(null);
+    }
```
Ben-Edwards-cgi, line 67: Why do we convert to BigInteger here if the data model only supports Long?

Either data model should be updated to be BigInteger or this method updated to long

src/main/java/uk/gov/hmcts/opal/service/opal/EnforcementAccountTypeService.java
```diff
@@ -30,6 +39,42 @@ public List<EnforcementAccountTypeCommon> getAllEnforcementAccountTypes() {
...
+    public List<EnforcementAccountTypeCommon> updateEnforcementAccountType(
+        List<PatchEnforcementAccountTypeRequestInner> request) {
+
+        checkPermissions();
+
+        List<EnforcementAccountTypeEntity> entities = new ArrayList<EnforcementAccountTypeEntity>();
+        for (PatchEnforcementAccountTypeRequestInner requestObject : request) {
+
+            EnforcementAccountTypeEntity entity =
+                repository.findById(requestObject.getId().longValue())
```
Ben-Edwards-cgi: The requestObject id should also be a long to align to the database value.

Else we won't be able to process all values if for example we had a account type with a very large id

src/main/java/uk/gov/hmcts/opal/service/opal/EnforcementAccountTypeService.java
```diff
@@ -30,6 +39,42 @@ public List<EnforcementAccountTypeCommon> getAllEnforcementAccountTypes() {
...
+            if (requestObject.getMinimumBalance() != null
+                && Optional.of(requestObject.getMinimumBalance()).orElse(new BigDecimal(0)).longValue() < 0L) {
+                throw new UnprocessableException("Can not set minimum balance to a negative value");
+            }
+
+            VersionUtils.verifyVersions(entity, BigInteger.valueOf(requestObject.getVersion().longValue()),
+                entity.getEnforcementAccountTypeId(), "updateEnforcementAccountType");
+
+            entity.setMinimumBalance(requestObject.getMinimumBalance());
+            entity.setVersionNumber(entity.getVersion().add(BigInteger.ONE).longValueExact());
```
Ben-Edwards-cgi: We should not manually set version numbers

Spring should do this automatically assuming we have the @Versioned annotation on the versioned field

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
```diff
@@ -73,4 +85,146 @@ public void getEnforcementAccountTypes_unauthorisedUser_shouldThrowPermissionsEx
...
+
+            verify(mapper).toEnforcementAccountTypeCommonList(List.of(mockEntity));
+        }
+    }
+
+    @Test
+    public void updateEnforcementAccountTypes_missingIdThrowsEntityNotFoundException() {
+        try (MockedStatic<SecurityUtil> securityUtil = mockStatic(SecurityUtil.class)) {
+            when(authToken.hasPermission(FinesPermission.AUTO_ENFORCEMENT)).thenReturn(true);
+            securityUtil.when(SecurityUtil::getOpalJwtAuthenticationTokenForCurrentUser).thenReturn(authToken);
```
Ben-Edwards-cgi: This is common setup logic could this be extracted into its own method?

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
```diff
@@ -73,4 +85,146 @@ public void getEnforcementAccountTypes_unauthorisedUser_shouldThrowPermissionsEx
...
+            List<PatchEnforcementAccountTypeRequestInner> request = List.of(
+                PatchEnforcementAccountTypeRequestInner.builder().id(1L).build()
+            );
+
+            when(repository.findById(eq(1L))).thenReturn(Optional.ofNullable(null));
```
Ben-Edwards-cgi: Optional: Instead of Optional.ofNullable(null))

it is best practice for Optional.empty()

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
```diff
@@ -73,4 +85,146 @@ public void getEnforcementAccountTypes_unauthorisedUser_shouldThrowPermissionsEx
...
+
+            assertThrows(EntityNotFoundException.class, () -> service.updateEnforcementAccountType(request));
```
Ben-Edwards-cgi: Need to validate the message of the error to ensure it is meaningful and contains the correct id

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
```diff
@@ -73,4 +85,146 @@ public void getEnforcementAccountTypes_unauthorisedUser_shouldThrowPermissionsEx
...
+    @Test
+    public void updateEnforcementAccountTypes_unauthorisedUser_shouldThrowPermissionsException() {
+        try (MockedStatic<SecurityUtil> securityUtil = mockStatic(SecurityUtil.class)) {
+            when(authToken.hasPermission(FinesPermission.AUTO_ENFORCEMENT)).thenReturn(false);
+            securityUtil.when(SecurityUtil::getOpalJwtAuthenticationTokenForCurrentUser).thenReturn(authToken);
+            List<PatchEnforcementAccountTypeRequestInner> request = List.of(
+                mock(PatchEnforcementAccountTypeRequestInner.class)
+            );
+
+            assertThrows(PermissionNotAllowedException.class, () -> service.updateEnforcementAccountType(request));
```
Ben-Edwards-cgi: Need to check error message

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
```diff
@@ -73,4 +85,146 @@ public void getEnforcementAccountTypes_unauthorisedUser_shouldThrowPermissionsEx
...
+            when(repository.findById(eq(1L))).thenReturn(Optional.of(
+                EnforcementAccountTypeEntity.builder()
+                    .enforcementAccountTypeId(1L)
+                    .enforcementAccountType(EnforcementAccountType.ADULT_NO_COLLECTION_ORDER_HIGH)
+                    .accountType(AccountType.ADULT_NO_COLLECTION_ORDER)
+                    .accountTypePath(LowHighValue.LOW)
+                    .build()
+            ));
+
+            assertThrows(UnprocessableException.class, () -> service.updateEnforcementAccountType(request));
```
Ben-Edwards-cgi: Need to check error message

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
```diff
@@ -73,4 +85,146 @@ public void getEnforcementAccountTypes_unauthorisedUser_shouldThrowPermissionsEx
...
+            );
+
+            when(repository.findById(eq(1L))).thenReturn(Optional.of(
+                EnforcementAccountTypeEntity.builder()
+                    .enforcementAccountTypeId(1L)
+                    .enforcementAccountType(EnforcementAccountType.ADULT_NO_COLLECTION_ORDER_HIGH)
+                    .accountType(AccountType.ADULT_NO_COLLECTION_ORDER)
+                    .accountTypePath(LowHighValue.LOW)
+                    .build()
+            ));
```
Ben-Edwards-cgi: This is common across a number of tests  could this move into a test data method?

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
```diff
@@ -73,4 +85,146 @@ public void getEnforcementAccountTypes_unauthorisedUser_shouldThrowPermissionsEx
...
+
+            assertThrows(UnprocessableException.class, () -> service.updateEnforcementAccountType(request));
```
Ben-Edwards-cgi: Need to check error message

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalEnforcementAccountTypesPatchIntegrationTest.java
```diff
@@ -0,0 +1,595 @@
...
+
+    @MockitoSpyBean
+    protected JsonSchemaValidationService jsonSchemaValidationService;
+
+    @MockitoBean
+    UserStateClientService userStateClientService;
+
+    @MockitoBean
+    protected AccessTokenService accessTokenService;
+
```
Ben-Edwards-cgi: We should not use mocks in integration tests

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalEnforcementAccountTypesPatchIntegrationTest.java
```diff
@@ -0,0 +1,595 @@
...
+    })
+    @Nested
+    class FeatureOn {
+
+        @Test
+        @DisplayName("OPAL: PATCH Enforcement Account Types - Update single enforcement account type")
+        @JiraStory("PO-2435")
+        @JiraEpic("PO-2433")
+        void patch_updatesSingleObjectOnly() throws Exception {
+            authorizeWithPermission((short) 78); // Auto enforcement permission
```
Ben-Edwards-cgi: We are moving away from ID based permissions (as these are not valid anymore).

As such we should use the permission name/ coded value instead

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalEnforcementAccountTypesPatchIntegrationTest.java
```diff
@@ -0,0 +1,595 @@
...
+            headers.add("Business-Unit-Id", "78");
+
+            String body = """
+                [
+                  {
+                    "id": 1,
+                    "version": 1,
+                    "minimum_balance": 200
+                  }
+                ]
```
Ben-Edwards-cgi: Instead of using JSON here can you use the DTO then use object mapper to convert it into JSON (This makes maintenance easier)

Same for all tests

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalEnforcementAccountTypesPatchIntegrationTest.java
```diff
@@ -0,0 +1,595 @@
...
+
+@ActiveProfiles({"integration", "opal"})
+@Slf4j(topic = "opal.OpalEnforcementAccountTypesPatchIntegrationTest")
+@Sql(
+    scripts = "classpath:db/deleteData/delete_from_enforcement_account_types.sql",
+    executionPhase = ExecutionPhase.AFTER_TEST_METHOD
+)
+class OpalEnforcementAccountTypesPatchIntegrationTest extends AbstractIntegrationWithSecurityTest {
+
+    protected static final String URL_BASE = "/enforcement-accounts-types/";
```
Ben-Edwards-cgi: No need for the end /

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalEnforcementAccountTypesPatchIntegrationTest.java
```diff
@@ -0,0 +1,595 @@
...
+                "SELECT * FROM enforcement_account_types WHERE enforcement_account_type_id = '1'",
+                eatRowMapper);
+            assertEquals(2L, firstChanged.getVersionNumber());
+            assertEquals(new BigDecimal("200.00"), firstChanged.getMinimumBalance());
+
+            EnforcementAccountTypeEntity secondChanged = jdbcTemplate.queryForObject(
+                "SELECT * FROM enforcement_account_types WHERE enforcement_account_type_id = '2'",
+                eatRowMapper);
+            assertEquals(2L, secondChanged.getVersionNumber());
+            assertEquals(new BigDecimal("300.00"), secondChanged.getMinimumBalance());
```
Ben-Edwards-cgi: Could this be extracted into its own method such as

assertEnforcementAccountType(id, version, amount)

thomas-croasdale-cgi: Done

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalEnforcementAccountTypesPatchIntegrationTest.java
thomas-croasdale-cgi: It appears we do need the trailing /
removing it results in 404 errors, and the openapi spec has it defined with the trailing /
https://github.com/hmcts/opal-fines-service/blob/master/src/main/resources/openapi/EnforcementAccountTypes.yaml#L12

Should I remove it from that spec?

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalEnforcementAccountTypesPatchIntegrationTest.java
thomas-croasdale-cgi: Done

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalEnforcementAccountTypesPatchIntegrationTest.java
thomas-croasdale-cgi: Changed the instances of this to not need the ID here. The AUTO_ENFORCEMENT permission is applied inside that function

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalEnforcementAccountTypesPatchIntegrationTest.java
thomas-croasdale-cgi: Removed

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
thomas-croasdale-cgi: Done

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
thomas-croasdale-cgi: Done

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
thomas-croasdale-cgi: Done

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
thomas-croasdale-cgi: Done

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
thomas-croasdale-cgi: Done

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
thomas-croasdale-cgi: Done

src/test/java/uk/gov/hmcts/opal/service/EnforcementAccountTypeServiceTest.java
thomas-croasdale-cgi: Done

src/main/java/uk/gov/hmcts/opal/service/opal/EnforcementAccountTypeService.java
thomas-croasdale-cgi: Both were already longs, the longValue() call was a mistake, this line is now updated to be correct

src/main/java/uk/gov/hmcts/opal/entity/enforcement/EnforcementAccountTypeEntity.java
thomas-croasdale-cgi: This is because the Versioned interface that we implement uses BigInteger instead of a long. Implementing that interface allows us to use the VersionUtils for verifyVersions().
imo if we want to change Versioned to use a long instead it should be raised as a tech debt, as it would have a bunch of side effects that I don't think should be covered by this PR

src/main/java/uk/gov/hmcts/opal/service/opal/EnforcementAccountTypeService.java
thomas-croasdale-cgi: Done

src/main/java/uk/gov/hmcts/opal/service/opal/EnforcementAccountTypesService.java
thomas-croasdale-cgi: This has been resolved in a later thread

src/main/java/uk/gov/hmcts/opal/entity/enforcement/EnforcementAccountTypeEntity.java
Ben-Edwards-cgi: Can you raise a Tech Debt and link here please then I am happy for this to be closed :)

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalEnforcementAccountTypesPatchIntegrationTest.java
Ben-Edwards-cgi: Good spot the spec is incorrect this trailing / should not be present please update

src/main/java/uk/gov/hmcts/opal/entity/enforcement/EnforcementAccountTypeEntity.java
thomas-croasdale-cgi: Created here: https://tools.hmcts.net/jira/browse/PO-9052

src/main/java/uk/gov/hmcts/opal/service/opal/EnforcementAccountTypeService.java
```diff
@@ -9,17 +15,22 @@
...
 import uk.gov.hmcts.opal.repository.EnforcementAccountTypeRepository;
+import uk.gov.hmcts.opal.util.VersionUtils;
 
 @Service
 @RequiredArgsConstructor
 public class EnforcementAccountTypeService {
 
     private final EnforcementAccountTypeRepository repository;
     private final EnforcementAccountTypeMapper mapper;
+    private final EntityManager entityManager;
```
Ben-Edwards-cgi: I don't think this is used so can be removed :)

thomas-croasdale-cgi: Done

Ben-Edwards-cgi (approved):

edoja1 (approved):
Looks good.

edoja1:
Approved

PR #1989, opened 2026-06-29
P0 2985-Get-Result-Add-Welsh-Language-Support

Elizabeth1420:
### 
- Added `include_welsh` support to `GET /results/{id}`
- Duplicates language-dependent text result parameters with a `cy_` prefix
- Sets Welsh duplicate hint to `Provide a welsh version for the defendant`
- Preserves existing behaviour when `include_welsh` is false or omitted ###

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/opal/service/opal/ResultService.java
+++ b/src/main/java/uk/gov/hmcts/opal/service/opal/ResultService.java
@@ -48,11 +48,7 @@ public class ResultService {
         return resultMapper.toRefData(getResultById(resultId));
     }
 
-    @Cacheable(value = "resultsCache", key = "#root.method.name + '_' + #resultId")
-    public ResultDto getResult(String resultId) {
-        return getResult(resultId, false);
-    }
-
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/opal/service/opal/ResultService.java
```diff
@@ -43,10 +50,53 @@ public ResultReferenceData getResultRefDataById(String resultId) {
```
Sean-Reed-CGI, line 50: This method doesn't appear to be used anymore. So maybe it should be removed.
Also should the @Cacheable be moved to the new method?

Elizabeth1420: I’ve removed the unused one-argument getResult method and moved @Cacheable onto the active getResult(resultId, includeWelsh) method.

I’ve also included includeWelsh in the cache key so Welsh and non-Welsh responses are cached separately.

Sean-Reed-CGI (approved):
LGTM

Alfred-Gillingham2 (approved):

PR #2003, opened 2026-07-01
Allow OA submitted-by IDs in functional response assertions

TimGonella:

N/A

### Change description ###

update functional response assertions to accept OA submitted-by user IDs alongside the existing JG variants
add a clarifying code comment explaining the JG to OA fallback for deployed environments
remove the old legacy/preprod gate so the assertion matches current environment behaviour

Why:
Demo functional tests were failing because responses returned account_snapshot.submitted_by values such as L073OA and L077OA, while the feature expectations still use L073JG and L077JG.

```
[ ] Yes
[ ] No
```


CadeFaulkner (approved):

ShanCGI (approved):
LGTM

PR #2006, opened 2026-07-02
PO-3951 Added read-only Feign retry policy for user service

Elizabeth1420:
### Implemented a read-only Feign retry policy for user service calls.

- Added ReadOnlyFeignRetryer to retry GET/HEAD requests only
- Scoped the retryer to the userService OpenFeign client via configuration
- Avoided retrying POST/write requests to prevent duplicate side effects
- Removed the global Retryer bean approach after it conflicted with pdpoSyncRetryer ###

```
[ ] Yes
[x] No
```


Elizabeth1420:
Closing this PR because the branch name contains parentheses, which breaks the Jenkins PR job shell command. Reopened from PO-3951-timeout-retry-handling.

PR #2009, opened 2026-07-03
PO-3832 - alter permission check logic for opal addNote

sarahtw-cgi:
https://tools.hmcts.net/jira/browse/PO-3832

### Change description ###
- Pass Business Unit Id in as a header parameter
- remove the account retrieval in NotesService
- get account from record_id in OpalNotesService
- Introduce dedicated `ADD_ACCOUNT_ACTIVITY_NOTES` fines permissions
- Update tests

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/opal/authorisation/model/FinesPermission.java
+++ b/src/main/java/uk/gov/hmcts/opal/authorisation/model/FinesPermission.java
@@ -20,6 +20,7 @@ public enum FinesPermission implements PermissionDescriptor {
     CHECK_VALIDATE_DRAFT_ACCOUNTS(5L, "Check and Validate Draft Accounts"),
     SEARCH_AND_VIEW_ACCOUNTS(6L, "Search and View Accounts"),
     ACCOUNT_MAINTENANCE(7L, "Account Maintenance"),
+    ADD_ACCOUNT_ACTIVITY_NOTES(8L, "Add account activity notes"),
     VIEW_CREDITOR_BACS(11L, "View Creditor BACS"),
     AMEND_PAYMENT_TERMS(9L, "Amend Payment Terms"),
     ENTER_ENFORCEMENT(10L, "Enter Enforcement"),
--- a/src/main/java/uk/gov/hmcts/opal/controllers/NotesController.java
+++ b/src/main/java/uk/gov/hmcts/opal/controllers/NotesController.java
@@ -32,14 +32,12 @@ public class NotesController {
     @Operation(summary = "adds a note to an entity")
     @FeatureToggle(feature = RELEASE_1B, defaultValueProperty = RELEASE_1B_ENABLED_PROPERTY)
     public ResponseEntity<String> addNote(
-        @RequestBody
-        AddNoteRequest request,
-        @RequestHeader("If-Match") String ifMatch) {
+        @RequestBody AddNoteRequest request,
+        @RequestHeader("If-Match") String ifMatch,
+        @RequestHeader("Business_Unit_ID") Short businessUnitId) {
 
         log.debug(":POST:postDefendantAccountSearch: query: \n{}", request.toPrettyJson());
-
-        String response =
-            notesService.addNote(request, ifMatch);
+        String response = notesService.addNote(request, ifMatch, businessUnitId);
 
         return buildCreatedResponse(response);
     }
--- a/src/main/java/uk/gov/hmcts/opal/service/legacy/LegacyNotesService.java
+++ b/src/main/java/uk/gov/hmcts/opal/service/legacy/LegacyNotesService.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.opal.service.legacy;
 
+import java.math.BigInteger;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
@@ -55,6 +56,6 @@ public class LegacyNotesService implements NotesServiceInterface {
             .recordId(request.getActivityNote().getRecordId()).build();
 
         return LegacyAddNoteRequest.builder().businessUnitId(defendantBusinessUnitId)
-            .businessUnitUserId(user.getUserId()).version(Long.valueOf(version)).activityNote(note).build();
+            .businessUnitUserId(user.getUserId()).version(new BigInteger(version)).activityNote(note).build();
     }
 }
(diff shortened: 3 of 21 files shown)


src/main/java/uk/gov/hmcts/opal/authorisation/model/FinesPermission.java
```diff
@@ -24,7 +24,8 @@ public enum FinesPermission implements PermissionDescriptor {
     AMEND_PAYMENT_TERMS(9L, "Amend Payment Terms"),
     ENTER_ENFORCEMENT(10L, "Enter Enforcement"),
     CONSOLIDATE(13L, "Consolidate"),
-    ADD_AND_REMOVE_PAYMENT_HOLD(14L, "Add and Remove payment hold");
+    ADD_AND_REMOVE_PAYMENT_HOLD(14L, "Add and Remove payment hold"),
+    ADD_ACCOUNT_ACTIVITY_NOTES(8L, "Add account activity notes");
```
RustyHMCTS: Could this be inserted in numerical order?

src/main/java/uk/gov/hmcts/opal/service/opal/OpalNotesService.java
```diff
@@ -70,4 +70,11 @@ public String addNote(AddNoteRequest req, String ifMatch, UserState user, Defend
         return entity.getNoteId().toString();
     }
```
Sean-Reed-CGI, line 60: Ideally the service layer shouldn't throw web specific exceptions

src/main/java/uk/gov/hmcts/opal/service/legacy/LegacyNotesService.java
```diff
@@ -21,13 +20,13 @@ public class LegacyNotesService implements NotesServiceInterface {
...
     @Override
-    public String addNote(AddNoteRequest request, String ifMatch, UserState user, DefendantAccountEntity account) {
+    public String addNote(AddNoteRequest request, String ifMatch, UserState user, Short businessUnitId) {
         log.info(":LegacyAddNote");
 
         GatewayService.Response<LegacyAddNoteResponse> response = gatewayService.postToGateway(
             ADD_NOTE,
             LegacyAddNoteResponse.class,
-            createRequest(request, ifMatch, user, account.getBusinessUnit().getBusinessUnitId().toString()),
+            createRequest(request, ifMatch, user, String.valueOf(businessUnitId)),
```
RustyHMCTS: I'm puzzled why we would need to convert this to a String like this. I suspect that the underlying LegacyAddNoteRequest class has the wrong types on it's properties.  Should they not be:
    
    @NotBlank
    @XmlElement(name = "business_unit_id", required = true)
    private Short businessUnitId;

    @NotBlank
    @XmlElement(name = "business_unit_user_id", required = true)
    private Long businessUnitUserId;

    @NotNull
    @XmlElement(name = "version", required = true)
    private BigInteger version;

@marcmollins - git tells me you were the original author?

src/main/java/uk/gov/hmcts/opal/service/opal/OpalNotesService.java
```diff
@@ -34,25 +36,23 @@ public class OpalNotesService implements NotesServiceInterface {
...
     @Override
     @Transactional
-    public String addNote(AddNoteRequest req, String ifMatch, UserState user, DefendantAccountEntity account) {
+    public String addNote(AddNoteRequest req, String ifMatch, UserState user, Short businessUnitId) {
         // TODO - waiting for PO-1564 to call DefendantAccountService to get the account
 
         log.info(":OpalAddNote");
 
-        Long accountId = account.getDefendantAccountId();
-        DefendantAccountEntity managed = em.find(DefendantAccountEntity.class, accountId);
```
RustyHMCTS, line 43: I believe we should not be using 'em' - which I see is EntityManager. I think some of the pre-existing code is problematic, and maybe we should try and address that now.

Sorry - this comment should have been added a few lines down.

RustyHMCTS (changes requested):

src/main/java/uk/gov/hmcts/opal/service/opal/OpalNotesService.java
marcmollins: I agree with this comment, this was before the project moved away from using entity manager. I think we should update this here (since the work is being done on this class) but maybe look to create a tech debt ticket to clean up other instances that still use Entity manager?

src/main/java/uk/gov/hmcts/opal/service/opal/OpalNotesService.java
sarahtw-cgi: Removed these from the service layer

src/main/java/uk/gov/hmcts/opal/service/legacy/LegacyNotesService.java
marcmollins: These would have came straight from the schemas and I believe back with these were written they were Strings. I can see the above changes have been made and happy to agree with above.

sarahtw-cgi: I have updated these

src/main/java/uk/gov/hmcts/opal/service/opal/OpalNotesService.java
sarahtw-cgi: I have removed the entity manager logic and tested that the method it's been replaced with still updates the version of the Defendant Account when notes is updated

RustyHMCTS (changes requested):
I think we still need to change version to BigInteger, as the versions returned from the Legacy system can be about 40 digits long. Search the code base for 'BigInteger version' to see where this is being used.

src/main/java/uk/gov/hmcts/opal/controllers/NotesController.java
```diff
@@ -32,14 +32,12 @@ public class NotesController {
...
+        @RequestHeader("Business_Unit_ID") Short businessUnitId) {
 
         log.debug(":POST:postDefendantAccountSearch: query: \n{}", request.toPrettyJson());
-
-        String response =
-            notesService.addNote(request, ifMatch);
+        String response = notesService.addNote(request, ifMatch, businessUnitId);
 
         return buildCreatedResponse(response);
     }
```
Sean-Reed-CGI, line 43: Missing unit test

src/test/java/uk/gov/hmcts/opal/service/OpalNotesServiceTest.java
```diff
@@ -0,0 +1,117 @@
+package uk.gov.hmcts.opal.service;
```
Sean-Reed-CGI: Wrong package

sarahtw-cgi: Thanks. Moved to live in the opal directory

src/main/java/uk/gov/hmcts/opal/controllers/NotesController.java
sarahtw-cgi: Added a success test

Sean-Reed-CGI (approved):
LGTM

RustyHMCTS (approved):

stephenmagill (approved):

PR #2011, opened 2026-07-03
Fix the integration Zephyr task wiring

TimGonella:
N/A

N/A

### Change description ###

This change only fixes the integration Zephyr task wiring: if the JUnit5 integration JSON is missing, the integration reporting step will fail for the real missing-file reason, but the Jenkins pipeline will still continue on to the functional stage because the integration stage error is caught and the build is marked failed rather than aborted

```
[ ] Yes
[ X] No
```


CadeFaulkner (approved):

marcmollins (approved):
LGTM

PR #2015, opened 2026-07-03
Po 2333 get defendant account consolidated accounts opal

HennaParmar:
https://tools.hmcts.net/jira/browse/PO-2333

### Change description ###
A new endpoint to retrieve accounts that were consolidated into this defendant account.

This API should use a newly created view "v_consolidated_accounts" to fetch data.

This API accepts a single path parameter which refers to the "master" defendant account for which the child accounts (returned apart of this API).

Master account = the account the children were consolidated into
Child account = Accounts that have been consolidated into the master

The user must have  'Search and View Accounts permission' in at least one business unit
If no child accounts are found an 200 with empty body should be returned

```
[ ] Yes
[ x ] No
```


src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalDefendantAccountConsolidatedAccountsIntegrationTest.java
```diff
@@ -0,0 +1,323 @@
...
+    private static final long EMPTY_MASTER_ACCOUNT_ID = 233304L;
+    private static final short BUSINESS_UNIT_ID = 78;
+    private static final short DIFFERENT_BUSINESS_UNIT_ID = 77;
+    private static final String URL = URL_BASE + "/%d/consolidated-accounts";
+
+    @MockitoBean
+    private UserStateService userStateService;
+
+    @BeforeEach
+    void setupConsolidatedAccountsData() {
```
alex-reeves-cgi, line 69: Should this be in a sql file?

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalDefendantAccountConsolidatedAccountsIntegrationTest.java
```diff
@@ -0,0 +1,323 @@
...
+
+        insertDefendantParty(CHILD_ACCOUNT_ID, "Alex", "Jones");
+        insertDefendantParty(OTHER_CHILD_ACCOUNT_ID, "Casey", "Smith");
+
+        insertConsolidationTransaction(23330001L, MASTER_ACCOUNT_ID, CHILD_ACCOUNT_ID);
+        insertConsolidationTransaction(23330002L, OTHER_MASTER_ACCOUNT_ID, OTHER_CHILD_ACCOUNT_ID);
+    }
+
+    @AfterEach
+    void deleteConsolidatedAccountsData() {
```
alex-reeves-cgi: Same as above

src/main/java/uk/gov/hmcts/opal/entity/defendantaccount/ConsolidatedAccountEntity.java
```diff
@@ -0,0 +1,50 @@
...
+@AllArgsConstructor
+public class ConsolidatedAccountEntity {
+
+    @Id
+    @Column(name = "master_account_id")
+    private Long masterAccountId;
+
+    @Id
+    @Column(name = "child_account_id")
+    private Long childAccountId;
```
alex-reeves-cgi, line 31: I had a look at the db and schema script, in the db this is a varchar(30) from the defendant_transaction.associated_record_id. I think this would only ever be a Long, but it's not enforced by the data. 

I don't know if we actually want to change this, it's just something to be aware of. If there were ever a data issue here it might be better if we deal with it rather than hibernate throwing (or could just pass it as a string in the dto?).

alex-reeves-cgi (commented):
Looks good, just the question about the int test data and the table column.

I also like the test display name with the int numbers, it's a great idea

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalDefendantAccountConsolidatedAccountsIntegrationTest.java
HennaParmar: Good point, I’ve moved the fixed integration setup data into src/integrationTest/resources/db/insertData/insert_into_consolidated_accounts.sql and wired it in with @Sql. The only data still created in Java is the generated bulk rows inside the large-result test, because that’s specific to that one scenario.

src/integrationTest/java/uk/gov/hmcts/opal/controllers/OpalDefendantAccountConsolidatedAccountsIntegrationTest.java
HennaParmar: Same as above -  I’ve moved the fixed integration setup data into src/integrationTest/resources/db/insertData/insert_into_consolidated_accounts.sql and wired it in with @Sql. The only data still created in Java is the generated bulk rows inside the large-result test, because that’s specific to that one scenario.

src/main/java/uk/gov/hmcts/opal/entity/defendantaccount/ConsolidatedAccountEntity.java
HennaParmar: Good spot. I’ve left the API field as numeric because the TDIA defines account_id as an integer and the view joins associated_record_id back to defendant_accounts.defendant_account_id, so only numeric values can be returned by this endpoint. I’ve also made the JPA identity composite on masterAccountId and childAccountId to avoid identity collisions.

alex-reeves-cgi (approved):
LGTM

DeclanClarkeCGI (approved):

vibhu-HMCTS (approved):

PR #2029, opened 2026-07-07
PO-8671 Remove Global Exception Handler class duplication in Fines Service and Common Lib

thanhdat-nguyen-hmcts:
https://tools.hmcts.net/jira/browse/PO-8671

### Change description ###
Removes duplicated shared exception handling from Fines Service and brings its error contract back in line with the common library.

- refactored `GlobalExceptionHandler` in Fines Service to rely on `opal-common-lib` for shared exception mappings
- removed duplicated local handlers and kept only the Fines-specific exception handling that still belongs in this service
- kept problem detail creation aligned with the shared common-lib behaviour
- restored a local high-precedence `EntityNotFoundException` handler to prevent generic 404 responses from leaking internal `reason` details
- updated unit and integration coverage for the sanitized generic not-found response contract

```
[ ] Yes
[x] No
```


Sean-Reed-CGI (approved):
LGTM

DeclanClarkeCGI (approved):

Alfred-Gillingham2 (approved):

PR #2047, opened 2026-07-13
PO-6449 update payment card request validate auth

ShanCGI:

https://tools.hmcts.net/jira/browse/PO-6449

### Change description ###

Add Payment Card Request for Defendant Account should evaluate if the user has permissions based on the authenticated user's userstate and the business-unit-id header parameter.

```
[ ] Yes
[x] No
```


Sean-Reed-CGI (approved):
LGTM

src/main/java/uk/gov/hmcts/opal/service/DefendantAccountPaymentTermsService.java
```diff
@@ -101,4 +98,11 @@ public GetDefendantAccountPaymentTermsResponse addPaymentTerms(Long defendantAcc
             throw new PermissionNotAllowedException(buId, FinesPermission.AMEND_PAYMENT_TERMS);
         }
     }
+
+    private String getBusinessUnitUserIdForBusinessUnit(UserState userState, short businessUnitId) {
+        return userState.getBusinessUnitUserForBusinessUnit(businessUnitId)
+            .map(BusinessUnitUser::getBusinessUnitUserId)
+            .filter(id -> !id.isBlank())
+            .orElseThrow(() -> new BusinessUnitUserNotFoundException(businessUnitId));
+    }
```
RustyHMCTS: I'm surprised we need functionality like this at this 'late' stage in the project - do we not have some similar functionality already in use?

src/main/java/uk/gov/hmcts/opal/controllers/advice/GlobalExceptionHandler.java
```diff
@@ -14,31 +18,89 @@
...
+            ex
+        );
+        problemDetail.setProperty("businessUnitId", ex.getBusinessUnitId());
+
+        return responseWithProblemDetail(HttpStatus.UNAUTHORIZED, problemDetail);
+    }
+
+    @ExceptionHandler(AccessDeniedException.class)
+    public ResponseEntity<ProblemDetail> handleAccessDeniedException(
+        AccessDeniedException ex, HttpServletRequest request) {
```
RustyHMCTS: Surprised we need these new exception handlers - can you just confirm that we're not replicating existing functionality that could be re-used? Thank.

RustyHMCTS (commented):
A couple of questions around whether we are potentially duplicating functionality. Thanks.

src/main/java/uk/gov/hmcts/opal/controllers/advice/GlobalExceptionHandler.java
ShanCGI: Refactored GloablExceptionHandler to use the existing `PermissionNotAllowedException` handler instead of `handleAccessDeniedException` / `handleBusinessUnitUserNotFoundException` handler advice.

`handleMissingRequestHeaderException()` - It doesn't look like handleMissingRequestHeaderException() is a duplicate. The other existing ecenarios are:

  - PermissionNotAllowedException: permission/auth failure, returns 403.
  - UnauthorizedException: invalid/missing token style failure, returns 401.
  - MissingMappingTypeException: domain-specific missing mapping type, returns 400 but different exception and problem type.

src/main/java/uk/gov/hmcts/opal/service/DefendantAccountPaymentTermsService.java
ShanCGI: This piece is duplicated in many of the calls or as a function in other classes like `DefendantAccountPartyService`, `DefendantAccountEnforcementService` etc. 

Have moved it as a common call to BusinessUnitService to be reused. Other calls can now use this.

RustyHMCTS (approved):

Ju-Morgan:
Review findings:
[P1]: Invalid Business-Unit-Id returns 500
Problem: Short.parseShort(businessUnitId) is called directly in DefendantAccountPaymentTermsService.java. Values such as abc or 40000 throw NumberFormatException, which is not mapped by GlobalExceptionHandler.
Why: This is still a client-supplied header validation failure, but it becomes a misleading 500—the failure mode PO-6449 is intended to eliminate. The OpenAPI describes the value merely as a string, so it also gives clients no constraint to prevent it.
Fix: Validate/map malformed or out-of-range IDs to a 400 problem response (and constrain the OpenAPI schema if appropriate). Add an integration test that asserts this response.

[P1]: No FINES-domain entry can still cause a 500 instead of 403
Problem: getDomainBusinessUnitUsers(Domain.FINES) may return null, then the permission check dereferences it at DefendantAccountPaymentTermsService.java.
Why: An authenticated user with no FINES business-unit state is not permitted for the selected BU and should receive the ticket’s required PermissionNotAllowedException → 403. The null-safe helper added in BusinessUnitService does not help, because execution fails before it is called.
Fix: Treat a null/empty FINES-domain BU collection as unauthorized for that BU and add a unit/integration case using a UserStateV2 with no FINES domain.

ShanCGI:
> Review findings: [P1]: Invalid Business-Unit-Id returns 500 Problem: Short.parseShort(businessUnitId) is called directly in DefendantAccountPaymentTermsService.java. Values such as abc or 40000 throw NumberFormatException, which is not mapped by GlobalExceptionHandler. Why: This is still a client-supplied header validation failure, but it becomes a misleading 500—the failure mode PO-6449 is intended to eliminate. The OpenAPI describes the value merely as a string, so it also gives clients no constraint to prevent it. Fix: Validate/map malformed or out-of-range IDs to a 400 problem response (and constrain the OpenAPI schema if appropriate). Add an integration test that asserts this response.

Updated add payment card path to align with controller pattern to handle invalid Business-Unit-Id

1. Controller binds Business-Unit-Id as Short
 
     For Business-Unit-Id: abc, Spring conversion fails before service logic.
      - Status: 406 Not Acceptable
      - Type: https://hmcts.gov.uk/problems/type-mismatch
      - Detail: Invalid parameter value format
      - Handler: common-lib OpalGlobalExceptionHandler.handleMethodArgumentTypeMismatchException(...)

ShanCGI:
> Review findings: 
> 
> [P1]: No FINES-domain entry can still cause a 500 instead of 403 Problem: getDomainBusinessUnitUsers(Domain.FINES) may return null, then the permission check dereferences it at DefendantAccountPaymentTermsService.java. Why: An authenticated user with no FINES business-unit state is not permitted for the selected BU and should receive the ticket’s required PermissionNotAllowedException → 403. The null-safe helper added in BusinessUnitService does not help, because execution fails before it is called. Fix: Treat a null/empty FINES-domain BU collection as unauthorized for that BU and add a unit/integration case using a UserStateV2 with no FINES domain.

Added null-safe check to handle missing FINES domain.

An authenticated user with no FINES business-unit state is not permitted for the selected BU and should receive the ticket’s required PermissionNotAllowedException → 403.

benji-telagalapudi76 (approved):

PR #2068, opened 2026-07-17
PO-8936 : Refactoring Defendant Account Code - Payment Terms

Sean-Reed-CGI:
https://tools.hmcts.net/jira/browse/PO-8936
"OpalDefendantAccountService refactor - Payment Terms Services"

This PR is focused Payment Terms code within the Defendant Account services.
It moves several methods from OpalDefendantAccountService to OpalDefendantAccountPaymentTermsService (and equivalent for legacy).
And removes some redundant code.

No functional changes.

### DIFF
--- a/src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountPaymentTermsService.java
+++ b/src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountPaymentTermsService.java
@@ -28,7 +28,7 @@ import uk.gov.hmcts.opal.util.VersionUtils;
 import java.time.LocalDate;
 
 @Service
-@Slf4j(topic = "opal.OpalDefendantAccountService")
+@Slf4j(topic = "opal.OpalDefendantAccountPaymentTermsService")
 @RequiredArgsConstructor
 public class OpalDefendantAccountPaymentTermsService implements DefendantAccountPaymentTermsServiceInterface {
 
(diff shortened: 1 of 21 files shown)


src/main/java/uk/gov/hmcts/opal/service/opal/OpalDefendantAccountPaymentTermsService.java
```diff
@@ -34,6 +40,22 @@ public class OpalDefendantAccountPaymentTermsService implements DefendantAccount
...
+    private final ReportEntryService reportEntryService;
+
+    private final PaymentTermsMapper paymentTermsMapper;
+
+    private final DefendantAccountRepository defendantAccountRepository;
+
+    private final EnforcementRepository enforcementRepository;
+
+    private final ResultService resultService;
+
```
RustyHMCTS, line 58: Just as a general observation, this seems like a lot of dependencies to bring in - has this class taken on too much responsibility?

Sean-Reed-CGI: OpalDefendantAccountService originally had 34 (!) dependencies before this refactoring started. Here we are moving code from ODAS into this class, so it's a step in the right direction.
But I completely agree this bean now has too many dependencies. I will raise a tech debt task to look at this issue separately from this ticket.

RustyHMCTS (approved):

PR #2038, opened 2026-07-09
PO-2574: Add new API endpoint to GET interface jobs summary.

RustyHMCTS:
Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #2084, opened 2026-07-21
Po 2578 test endpoint delete interface jobs

alex-reeves-cgi:

https://tools.hmcts.net/jira/browse/PO-2578

### Change description ###

This adds a testing support endpoint to DELETE interface jobs (give a list of ids), and all interface messages, interface files, tills and payments_in associated with the job.

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/opal/entity/InterfaceFileEntity.java
```diff
@@ -0,0 +1,62 @@
+package uk.gov.hmcts.opal.entity;
+
```
Sean-Reed-CGI, line 2: These entities are also being created in https://github.com/hmcts/opal-fines-service/pull/2038 and that PR is ahead of this one. 
So perhaps this ticket is blocked until that goes in?

Sean-Reed-CGI (approved):
LGTM

src/main/java/uk/gov/hmcts/opal/entity/InterfaceFileEntity.java
alex-reeves-cgi: good spot, I agree

TimGonella (dismissed):
[P1] ids is treated as optional, but the ticket requires it Problem: The controller accepts ids as required = false, and the service returns 200 OK when the list is null or empty. See src/main/java/uk/gov/hmcts/opal/controllers/TestingSupportController.java:100 and src/main/java/uk/gov/hmcts/opal/service/opal/InterfaceJobService.java:111. Why: The story’s AC and INT.04 require ids to be mandatory and well-formed, with invalid requests rejected via the standard error handling. The new integration test also bakes in the wrong behaviour by asserting 200 for no ids at src/integrationTest/java/uk/gov/hmcts/opal/controllers/TestingSupportControllerDeleteInterfaceJobsIntegrationTest.java:98. Fix: Make ids required, validate non-empty numeric input at the controller boundary, and change the integration coverage to expect the documented 4xx response.

[P1] The testing endpoint is not hidden as required Problem: The new delete method is feature-flagged, but it is not annotated with @Hidden. See src/main/java/uk/gov/hmcts/opal/controllers/TestingSupportController.java:93. Why: The ticket explicitly requires this endpoint to be hidden and disabled in production using @Hidden plus @ConditionalOnProperty. Other testing-only endpoints in this repo follow that pattern, e.g. src/main/java/uk/gov/hmcts/opal/controllers/DraftAccountController.java:112 and src/main/java/uk/gov/hmcts/opal/controllers/MinorCreditorController.java:75. Fix: Add @Hidden to the delete endpoint and add a test that proves the endpoint is absent when opal.testing-support-endpoints.enabled=false.

[P2] The integration coverage does not match the ticketed scenarios Problem: The new integration class only covers happy path, empty ids, and a forced DB failure. It does not cover property-disabled 404, required headers/validation, documented 200 schema, missing/non-existent ids behavior, repeat delete semantics, or non-OPAL/production safety. See src/integrationTest/java/uk/gov/hmcts/opal/controllers/TestingSupportControllerDeleteInterfaceJobsIntegrationTest.java:37. Why: The story lists INT.03–INT.09 explicitly, and the integration-test review skill expects the tests to prove the intended backend behaviour rather than only a subset. Fix: Add the missing controller integration cases, and give each test method the required @DisplayName values; the current methods do not have them.

Typo in the new step class name (Interfece...)

benji-telagalapudi76 (approved):

benji-telagalapudi76 (approved):

jonathanDuffy (approved):

PR #2127, opened 2026-07-31
PO-8973: Migrate get results by id

ks21285-hmcts:

https://tools.hmcts.net/jira/browse/PO-8973

### Change description ###
Migrates the manually implemented GET /results/{id} endpoint to use the OpenAPI contract defined in Results.yaml.
It preserves the endpoint; it replaces the handwritten controller contract and DTO with generated OpenAPI classes.

```
[ ] Yes
[x] No
```


alex-reeves-cgi (approved):
Looks good, note that this is also including https://tools.hmcts.net/jira/browse/PO-8972, so you should update that ticket too.  I am a little concerned that it didn't include the getResults (search results?) so the logic of results are split over multiple controllers, do you know if there is a ticket for this?

src/main/resources/openapi/Results.yaml
```diff
@@ -0,0 +1,156 @@
...
+        '503':
+          description: Service unavailable
+          content:
+            application/problem+json:
+              schema:
+                $ref: './common.yaml#/components/schemas/ProblemDetail'
+
+components:
+  schemas:
+    GetResultByIdResponse:
```
ShanCGI, line 70: **Problem**: The new generated response model is produced from `Results.yaml` with the repository template's class-level `JsonInclude.Include.ALWAYS`. Consequently, every optional field is now serialised as `null`. The deleted `ResultDto` used `JsonInclude.Include.NON_NULL` for the response as a whole and overrode only `allow_additional_action` to `ALWAYS`, so fields such as `imposition_category`, `result_parameters`, and `requires_lja` were previously omitted when null.

Existing consumers will receive a different JSON shape for results containing nullable database values, which can break strict clients or presence-based logic.

**Fix**: would suggest configuring this generated schema/model to omit null-valued optional properties while continuing to emit `allow_additional_action: null` (for example, through scoped generator/vendor annotations or an appropriate template change). Add a serialisation test covering both behaviours and an endpoint-level test using a result whose optional fields are null.

ks21285-hmcts:
@alex-reeves-cgi , thanks for your comment. I could not reply above as it was already resolved. 
I will include the ticket you mentioned. 
Yes, I noticed that there is a missing ticket for just getResults(), @DeclanClarkeCGI  recommended creating one, which I will create. Once that is done all the results will be in the same controller.

src/main/resources/openapi/Results.yaml
ks21285-hmcts: Good point, I have modified moustache such a way that we can now extra tag to have NON_NULL and ALWAYS.
Added unit and integration tests to check null and optional fields.

ShanCGI: LGTM

ShanCGI (approved):
LGTM

DeclanClarkeCGI (approved):

TimGonella (approved):

PR #2140, opened 2026-08-06
PO-2452-Using-Optimistic-Lock-And-Automatic-Update

Elizabeth1420:
### Change description ###
Replaced the non-atomic in-memory `versionNumber + 1` logic in `bumpVersion` with an atomic repository update.

Uses the existing account loaded by `findById` for validation and mutation
- flushes pending account-party changes before bumping the version
- increments `defendant_accounts.version_number` only when the current DB version matches the expected version
- returns the newly stored version so the response `ETag` reflects the bumped version immediately
- throws a conflict if the version was changed by another transaction before the bump

Tests:
 Updated `OpalDefendantAccountServiceTest03` to verify `incrementVersionNumber(...)` is called during successful replace-party flows
-  Added `DefendantAccountRepositoryServiceTest` coverage for:
- successful atomic version increment returning the updated version
- conflict handling when no row is updated

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/opal/repository/DefendantAccountRepository.java
+++ b/src/main/java/uk/gov/hmcts/opal/repository/DefendantAccountRepository.java
@@ -6,6 +6,7 @@ import java.util.Optional;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
 import org.springframework.data.jpa.repository.Lock;
+import org.springframework.data.jpa.repository.Modifying;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.query.Param;
 import org.springframework.stereotype.Repository;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/opal/repository/DefendantAccountRepository.java
```diff
@@ -36,4 +37,16 @@ Optional<DefendantAccountVersionData> findVersionDataByDefendantAccountId(
...
+    @Modifying
+    @Query("""
+        UPDATE DefendantAccountEntity d
+        SET d.versionNumber = d.versionNumber + 1
+        WHERE d.defendantAccountId = :defendantAccountId
+        AND d.versionNumber = :expectedVersion
+        """)
+    int incrementVersionNumber(
+        @Param("defendantAccountId") Long defendantAccountId,
+        @Param("expectedVersion") Long expectedVersion
```
ks21285-hmcts, line 49: Did `expectedVersion` mean existingVersion?

Elizabeth1420: Yes it does.

ks21285-hmcts (approved):
LGTM

src/main/java/uk/gov/hmcts/opal/repository/DefendantAccountRepository.java
```diff
@@ -36,4 +37,16 @@ Optional<DefendantAccountVersionData> findVersionDataByDefendantAccountId(
     @Query("select d from DefendantAccountEntity d where d.defendantAccountId = :id")
     Optional<DefendantAccountEntity> findByDefendantAccountIdForUpdate(@Param("id") Long defendantAccountId);
```
Sean-Reed-CGI, line 39: Did you consider using:
  `entityManager.lock(defendantAccountiEntity, LockModeType.OPTIMISTIC_FORCE_INCREMENT);`
instead of manually updating the version?

Elizabeth1420: Yes, I tried the approach first.

The issue was that although it can force the version increment, the bumped version was not visible early enough for this endpoint’s response. The integration tests showed the response `ETag` was still returning the current version instead of `currentVersion + 1`.

This endpoint needs to return the newly bumped version immediately in the response, so I used an atomic DB update with the expected version in the `WHERE` clause. That still gives us optimistic-lock behaviour because the update only succeeds if the DB version matches the version we originally read. If another transaction has already changed it, no row is updated and we return a conflict.

I also added/updated tests around this, including the `DefendantPartyPutIntegrationTest`, which now passes and confirms the `ETag` returns the bumped version.

Ant-Collins (approved):

PR #2167, opened 2026-08-14
PO 1858 Account status date not updated after patching a new status of deleted

HennaParmar:
https://tools.hmcts.net/jira/browse/PO-1858

### Change description ###
When a status of "Deleted" is patched to an existing draft account using the draft-accounts endpoint the account_status_date is not being updated to reflect the deleted date. The PATCH flow was already updating the account status to Deleted, but it was not updating account_status_date. The fix adds that timestamp update when the deleted status is applied.
This should resolve the issue where the Deleted tab shows the wrong date for deleted draft accounts.

```
[ ] Yes
[ X ] No
```


RustyHMCTS (approved):

edoja1 (approved):
Approved

PR #2202, opened 2026-08-25
Po 5762 update user state v2 to change from numberic i ds to coded value enum name fines

AlanRogers-user:

https://tools.hmcts.net/jira/browse/PO-5762

### Change description ###
Updated to use the new UserStateV2, BusinessUnitUserV2 and PermissionV2 instances.

All tests have been updated to use these new class instances.

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/opal/authorisation/model/FinesPermission.java
```diff
@@ -2,43 +2,43 @@
...
+    ENTER_ENFORCEMENT("ENTER_ENFORCEMENT", "Enter Enforcement"),
+    CONSOLIDATE("CONSOLIDATE", "Consolidate"),
     // TODO verify this ID mirrors opal-user-service Permissions.ADD_AND_REMOVE_PAYMENT_HOLD ?
-    ADD_AND_REMOVE_PAYMENT_HOLD(14L, "Add and Remove payment hold"), // TODO - should this be 12L?
-    PROCESS_AND_ALLOCATE_PAYMENTS(16L, "Process and Allocate Payments"),
-    AUTO_ENFORCEMENT(17L, "Auto Enforcement");
+    ADD_AND_REMOVE_PAYMENT_HOLD("ADD_AND_REMOVE_PAYMENT_HOLD", "Add and Remove payment hold"),
+    PROCESS_AND_ALLOCATE_PAYMENTS("PROCESS_AND_ALLOCATE_PAYMENTS", "Process and Allocate Payments"),
+    AUTO_ENFORCEMENT("AUTO_ENFORCEMENT", "Auto Enforcement"),
+    DRAFT_ACCOUNT_PERMISSIONS("DRAFT_ACCOUNT_PERMISSIONS", "Draft Account Permissions");
```
HennaParmar: Do we need DRAFT_ACCOUNT_PERMISSIONS as an enum value? It looks like this was previously only a convenience grouping of the two draft permissions, and draftAccountPermissions() still provides that grouping. Adding it to the enum also means it will now be included in FinesPermission.values() as if it were a real permission.

src/integrationTest/java/uk/gov/hmcts/opal/support/UserStateStub.java
```diff
@@ -102,12 +108,59 @@ private static String generateJwt() {
 
     public String getUserStateAsJson() {
         try {
+            UserStateV2 userState = getUserState();                 //  Temp
```
HennaParmar: Are these temporary serialisation/deserialisation lines still needed? userState, userStateAsJson and decoded don't appear to be used, and the method calls getUserState() again for the return value.

src/integrationTest/java/uk/gov/hmcts/opal/support/UserStateStub.java
```diff
@@ -245,22 +298,19 @@ public UserStateV2Builder getDefaultUserStateBuilder() {
...
             .collect(Collectors.toSet());
     }
 
-    private Permission createPermission(PermissionDescriptor permissionDescriptor) {
-        return Permission.builder()
-            .permissionId(permissionDescriptor.getId())
-            .permissionName(permissionDescriptor.getDescription())
-            .build();
+    private PermissionV2 createPermission(PermissionDescriptorV2 permissionDescriptor) {
+        return PermissionV2.fromPermissionName(permissionDescriptor.getPermissionName());
```
HennaParmar: Should this map via permissionCode/fromPermissionCode instead? FinesPermission.toCommonPermission() now uses the code, which seems preferable to relying on the display name for identity.

src/main/java/uk/gov/hmcts/opal/authorisation/model/FinesPermission.java
AlanRogers-user: Fair comment.
I will address that.

src/integrationTest/java/uk/gov/hmcts/opal/support/UserStateStub.java
AlanRogers-user: No. Sorry.. I just forgot to remove it!

src/main/java/uk/gov/hmcts/opal/authorisation/model/FinesPermission.java
AlanRogers-user: Done!

src/integrationTest/java/uk/gov/hmcts/opal/support/UserStateStub.java
AlanRogers-user: Fixed!

src/integrationTest/java/uk/gov/hmcts/opal/support/UserStateStub.java
AlanRogers-user: I think that this needs some thinking to resolve.

AlanRogers-user: Thios proved to be simpler than I thought.

HennaParmar (approved):

PR #2218, opened 2026-08-26
Po 9152 defendant account balance flip

alex-reeves-cgi:

https://tools.hmcts.net/jira/browse/PO-9152

### Change description ###

Legacy stores and produces a negative value for defendant_accounts' amount_imposed and account_balance, only the frontend should display a non negative value - and that is what we are keeping (although I personally disagree). 

The validator only uses the account balance, but as both values should be negative I've updated that in the tests which was the majority of the development.

Also, this ticket is reliant on the opal-legacy-db-stub PR https://github.com/hmcts/opal-legacy-db-stub/pull/304. It is a small change to test data for LegacyDefendantsCommonIntTest, attached is a screenshot that shows the failing test passes when using the new stub.
image

```
[ ] Yes
[x] No
```


sarahtw-cgi (approved):

Ju-Morgan (approved):
ALGTM

PR #2227, opened 2026-08-27
PO-8990 add notes to open api

alex-reeves-cgi:

https://tools.hmcts.net/jira/browse/PO-8990

### Change description ###

moved POST add notes endpoint to use open API 

It has been brought up that the path to the add notes endpoint should be /notes, however the old endpoint was notes/add, therefore the path is different and so FE may need a fix.

```
[X] Yes
[] No
```


src/integrationTest/java/uk/gov/hmcts/opal/controllers/shared/NotesIntegrationTest.java
```diff
@@ -40,18 +40,21 @@ void setUp() {
...
+                .build())
+            .build();
 
-        AddNoteRequest request = new AddNoteRequest();
-        request.setActivityNote(note);
 
         final String payload = objectMapper.writeValueAsString(request);
         log.info(":testPostNotes payload: {}", payload);
 
+        log.info(":  alt request payload: {}", request.toJsonString());
```
sarahtw-cgi: are both of these log infos needed?

sarahtw-cgi (commented):
LGTM. One small question on logging. 
Also ran locally against frontend and confirmed it does not break the notes functionality.

sarahtw-cgi (approved):

src/main/resources/openapi/notes.yaml
```diff
@@ -0,0 +1,112 @@
+openapi: 3.1.1
+info:
+  title: Opal Fines Service - Notes
+  version: 1.0.0
+  description: APIs for managing notes.
+paths:
+  /notes/add:
```
Ben-Edwards-cgi: This does not align to the logical API spec:
https://tools.hmcts.net/confluence/spaces/PO/pages/1864467433/Add+Note

Please review and ensure this spec aligns to what is outlined in the logical

src/main/resources/openapi/notes.yaml
```diff
@@ -0,0 +1,112 @@
...
+          description: Version of the defendant account. DB Mapping - defendant_accounts.version
+          required: true
+          schema:
+            type: string
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/AddNoteRequest'
```
Ben-Edwards-cgi, line 32: Should reference a common component in accordance to the logical spec

src/main/resources/openapi/notes.yaml
```diff
@@ -0,0 +1,112 @@
...
+            - creditor_accounts
+            - suspense_accounts
+            - report_instances
+        record_id:
+          type: string
+        note_text:
+          type: string
+        note_type:
+          type: string
+          const: AA
```
Ben-Edwards-cgi: Needs restrictions on min/max length in alignment with the DB

src/main/resources/openapi/notes.yaml
alex-reeves-cgi: There is a common component at https://github.com/hmcts/opal-fines-service/blob/2b594325128116d3203271c56c1bd45f2d74b0c1/src/main/resources/openapi/common.yaml#L254 that is similar, but the enum values do not match the previous values (or the ones in the common object defs). I could update that component in the yaml and use that $ref, although it's a bit scary?

src/main/resources/openapi/notes.yaml
alex-reeves-cgi: Ext. comment above, this would be an edit to the common component? I think only record_id has a fixed length at 30, I can add that

src/main/resources/openapi/notes.yaml
Ben-Edwards-cgi: Can you check with the FE to assess the impact, as ideally we should be using the common components to align to the logical spec just so if we update the common it cascades correctly.

src/main/resources/openapi/notes.yaml
Ben-Edwards-cgi: As the database has fixed width I think this is a low risk change as it will be impossible for a value larger or smaller then the database lengths to be retrieved / stored without error.

Adding this into the spec gives nicer errors then generic 500

src/main/resources/openapi/notes.yaml
alex-reeves-cgi: I've checked with FE, they're doing the enum to the spec, but the path to BE so has the /add. This is using the common component now, this is currently the only usage of it and is now to spec

Ben-Edwards-cgi: What do you mean by the /add sorry?

If the FE is calling POST /notes/add we will need a FE ticket to be played and merged alongside this one to use the correct URL

src/main/resources/openapi/common.yaml
```diff
@@ -259,12 +259,13 @@ components:
...
           type: string
           enum:
-            - Defendant Account
-            - Creditor Account
-            - Suspense Account
-            - Report Instances
+            - defendant_accounts
+            - creditor_accounts
+            - suspense_accounts
+            - report_instances
```
Ben-Edwards-cgi, line 265: Can you confirm with the FE to assess the impact of this change please.

We should be using the logical spec but we will need them to review all usages of this enum before we apply this change (else FE will break)

alex-reeves-cgi:
DO NOT MERGE: FE ticket https://tools.hmcts.net/jira/browse/PO-10535 needs to be ready to promote at the same time, as these changes will introduce breaking changes on the FE.

src/main/resources/openapi/notes.yaml
alex-reeves-cgi: Your other comment linking to the logical API spec on notes.yaml:7

alex-reeves-cgi:
Questions about FE impact: we have https://github.com/hmcts/opal-frontend/pull/3082 PR that addresses those and will be merged alongside this PR

Ben-Edwards-cgi (approved):
Approved from admin perspective.

Note: I have not reviewed any code which does not require admin review as such another dev will be required to do this

PR #2281, opened 2026-09-10
PO-8948: Create add tills OpenAPI schema

shaunw-cgi:
https://tools.hmcts.net/jira/browse/PO-8948

### Change description ###
Create add till Open API schema

```
[ ] Yes
[x] No
```


ks21285-hmcts (approved):
LGTM

Nipa1410 (approved):

src/main/resources/openapi/Tills.yaml
```diff
@@ -142,6 +208,201 @@ paths:
 
 components:
   schemas:
+    TillsCreateRequest:
+      type: object
+      additionalProperties: false
+      required:
+        - business_unit_id
+        - payments_in
+      properties:
```
marcmollins, line 217: the add tills page has both:

Payments In | 1..1

Payment In | 1..N

I think there should be an object above Payment_in?

marcmollins (changes requested):

src/main/resources/openapi/Tills.yaml
shaunw-cgi: I read Payments In as the array and Payment In as each item within the array, rather than separate JSON properties. This matches the API paths, for example `$.payments_in[].defendant_account_id`. Would adding another payment_in object change the payload path to `$.payments_in.payment_in[].defendant_account_id`, differing from the current spec?

marcmollins (approved):
LGTM

PR #2315, opened 2026-09-21
PO-3630: Create Add Till API

shaunw-cgi:
https://hmcts.atlassian.net/browse/PO-3630

### Change description ###
- Adds the feature-flagged POST /tills endpoint for creating manual payment tills.
- Persists the till and linked payments atomically, with business-unit permission validation.
- Triggers the pre-allocated Cash Till report and rolls back if report creation fails.
- Adds unit and integration coverage for creation, validation, authorisation, feature flag, and rollback behaviour.

```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/opal/controllers/r1c/CreateTillIntegrationTest.java
```diff
@@ -0,0 +1,289 @@
...
+
+    @Test
+    @JiraStory("PO-3630")
+    @JiraEpic("PO-2439")
+    void postTills_rejectsRequestWithoutBusinessUnitId() throws Exception {
+        mockMvc.perform(post(URL)
+                .with(userStateStub.getAuthenticaitonRequestPostProcessor())
+                .contentType(MediaType.APPLICATION_JSON)
+                .content("{\"payments_in\":[]}"))
+            .andExpect(status().isBadRequest());
```
sarahtw-cgi, line 102: This doesn't explicitly check that it's the missing BU which is causing the bad request. Could you also assert against the error message? Same for the rest of the bad request tests

ks21285-hmcts (commented):
LGTM
It would be nice to add a bruno request

src/main/java/uk/gov/hmcts/opal/mapper/till/CreateTillMapper.java
```diff
@@ -0,0 +1,61 @@
...
+                            BigDecimal totalAmount, Short paymentsCount, LocalDateTime createdDate);
+
+    @Mapping(target = "paymentInId", ignore = true)
+    @Mapping(target = "tillEntity", source = "till")
+    @Mapping(target = "paymentAmount", source = "requestPayment.paymentDetails.amount")
+    @Mapping(target = "paymentMethod", source = "requestPayment.paymentDetails.method")
+    @Mapping(target = "destinationType", source = "requestPayment.paymentDetails.destinationType")
+    @Mapping(target = "allocationType", source = "requestPayment.paymentDetails.allocationType")
+    @Mapping(target = "thirdPartyPayerName", source = "requestPayment.paymentDetails.thirdPartyPayerName")
+    @Mapping(target = "receipt", constant = "false")
```
sarahtw-cgi, line 55: Should these be hard-coded to false?

sarahtw-cgi (commented):
Looks good. A couple of comments. 

I saw your comment in the ticket. Have you had confirmation that `date` and `request` aren't needed and didn't need to be added to the spec and the implementation?

Also has the LLD been updated?

src/main/java/uk/gov/hmcts/opal/mapper/till/CreateTillMapper.java
shaunw-cgi: I have updated receipt and allocated so the mapper ignores their default state as they are not request fields. I have left autoPayment=false as this mapper is used for manual till payment creation

src/integrationTest/java/uk/gov/hmcts/opal/controllers/r1c/CreateTillIntegrationTest.java
shaunw-cgi: You're right, I've updated the missing BU test so it removes only the business_unit_id and added assertions for the problem detail messages across all bad requests

shaunw-cgi:
> Looks good. A couple of comments.
> 
> I saw your comment in the ticket. Have you had confirmation that `date` and `request` aren't needed and didn't need to be added to the spec and the implementation?
> 
> Also has the LLD been updated?

I've had confirmation that the API Spec is the source of truth and the requirements for `date` and `request` from the ticket are outdated and no longer required. LLD has been updated.

sarahtw-cgi (approved):

